Issue #9, opened 2021-05-15
Data structures 


Jeadie:
Hi Ramnes, 

I love the work you've done so far on the Notion SDK for python. I had a couple ideas to implement in the SDK if you thought they'd be of value: 

- Simple data classes on top of the various API objects for Typing and abstraction. 
- Implementing natural pythonic pagination and iteration 

I'll fork the repo and start work on these. Let me know what you think of these and I can make a PR when I'm done.

Cheers, 
Jeadie

ramnes:
Yeah I really want this too, but I'm torn between keeping the library API as close as possible to the javascript SDK, or making it higher level, i.e. in the lines of notion-py. I was thinking that maybe we could keep this library quite bare and write another one based on it, with much more batteries included, just like botocore and boto3 for example. What do you think?

1g0rrr:
But the original library has "api-types.ts". Do you plan to implement something similar?

> Yeah I really want this too, but I'm torn between keeping the library API as close as possible to the javascript SDK, or making it higher level, i.e. in the lines of notion-py. I was thinking that maybe we could keep this library quite bare and write another one based on it, with much more batteries included, just like botocore and boto3 for example. What do you think?

Jeadie:
I can't see that hurting, especially to keep more minimal python integrations, for example, less bloated. I'll start a new repo tomorrow (GMT+10) and use your library as the underlying connection.

Blue-Hope:
btw I agree with @ramnes that we should develop further, not stay in Js SDK. The api level we support should follow the official api announcement, but we will be able to support additional functions such as watching record in notion-py. The priority is to follow our library as much as possible, similar to js :)

ramnes:
Yeah I think we all want a higher level experience. The only questions are what features, what API, and how to structure the code.

After more thoughts, I'm wondering if the double-package scheme is optimal, and if we shouldn't rather have everything in here with different Python namespaces, so that the switch to a code generator would be easier when they'll release an OpenAPI specification file. That would also make more people work on a single codebase, which is a good way to increase the overall bus factor.

Maybe we could put all the current code in a `notion_client.base` module (or "sdk", or "core"), which would always stay the equivalent of the JS SDK, and write some nice wrappers around it in the first-level package, just like I've made for [sftpgo-client](https://github.com/ramnes/sftpgo-client).

What do you guys think?

Jeadie:
[...]

PR #10, opened 2021-05-15
[WIP] Dataclasses for API objects.

Jeadie:
As mentioned in this [issue](https://github.com/ramnes/notion-sdk-py/issues/9), work towards creating pythonic access to api objects.


aahnik:
this is an awesome pr. everyone needs pythonic access. looking forward to this getting merged.

ramnes:
Can you rebase on top of main? You have a lot of unrelated commits included here and it makes quite some noise.

Jeadie:
Oops, I didn't notice this when working on my fork. Ill be sure to clean it all up properly before asking to be approved.

ramnes:
Hey @Jeadie, thanks for the work so far, it looks quite good. I'm seeing that you're beginning to add more and more stuff though, can we try to keep this PR atomic? It will be hard to review and get merged in otherwise. More precisely, I think your last commit should be in another PR, so that we can focus on the data types here.

Jeadie:
Alright so a first draft of the dataclasses is up. For some reason we are using Codecov as a GH action but not writing tests. I'll move the remaining work to a new issue in my repository so I can keep track of them.

aahnik:
this is a giant pr. why don't some things get merged, and more work can be continued in upcoming prs.

what are the plans with the `lib` folder? (I think I saw some discussion about it. was those comments deleted ? i could not find them anymore)

how will things be structured finally?

ramnes:
> Alright so a first draft of the dataclasses is up. 

Great, thanks for work! It definitely looks like a big one. :)

Can you modify examples to make the value of this PR more clear? At the moment it doesn't seem usable as is to me; did you only plan to leverage these types with your `Notion` object?

Also, one thing that I'm concerned with is the description of every little thing. Shouldn't we make the inference a bit more generic so that we don't break for every minor change of Notion's API? I feel that we can have something that brings as much value to the library user without being so strict and prompt to breakage.

Also, commits are still a bit messed up, there are commits that shouldn't be there (i.e. the commits that you didn't write yourself).

> For some reason we are using Codecov as a GH action but not writing tests.

Good point, we should probably write the tests before merging this. I think @Blue-Hope is working on this (?) but I'll probably give it a try as well.

> how will things be structured finally?

I think I'd prefer the current code to be moved to a `notion_client.base` directory those other things implemented in `notion_client`. What do you think?

aahnik:
>I think I'd prefer the current code to be moved to a notion_client.base directory those other things implemented in notion_client. What do you think?

That's great. 

We also need to update the tests, examples, and documentation accordingly.

Another question, is it possible to make a pr to a pr?

aahnik:
> we should probably write the tests before merging this

We can only write tests for the base. After merging this, the import paths need to be amended. If anyone is not working on tests for the base, I am up for it.

We can write tests for the high-level functions and objects defined in this PR, only after this is merged.

notion_client/lib/custom_enums.py
```diff
@@ -0,0 +1,124 @@
+from enum import Enum
+
+
+class BasicColor(Enum):
```
aahnik:
```suggestion
class BasicColor(str,Enum):
```

Using str based enums is better for all the Enums in this module.

```python
from enum import Enum


class MyEnum(Enum):
    state1 = "state1"
    state2 = "state2"


assert MyEnum.state1 == "state1" # will raise error

```

The error can be fixed by explicitly defining the type

```python
class MyEnum(str,Enum)
```

ramnes: https://docs.python.org/3/library/enum.html#others

TIL

aahnik: "TIL"

TIL

jheddings:
Has there been any thoughts or discussion about how this will evolve to support writing back to Notion?  It looks like a great wrapper for reading data, but I'm not seeing how it will support writing data without a lot of effort.  It could be there and I am missing it, just asking in case it has any bearing on the outcome.

nehiljain:
> Another question, is it possible to make a pr to a pr? I would love to change small things here.

+1 
Great work @Jeadie !!
I would love to do the same. It feels like breaking the work and the size of the PR is the way to remove the bottleneck. Please let me know how I can start contributing

aahnik:
Hi @Jeadie when will this pr ship? I am ultra excited.

notion_client/lib/api_objects.py
```diff
@@ -0,0 +1,430 @@
...
+    print(d)
+    property_value: PropertyValue = property_type_to_class[property_type].from_json(d)
+
+    print(property_value)
+    print()
+    return property_value
+
+
+@dataclass
+class PropertyValue:
```
MrCredible, line 131: Wouldnt propertyValue be a subclass of the class Property?

The property values that pages have provide all the same information as the properties objects within databases except they have actual value information.

By not doing so, it looks like it may lead to double maintenance.

Example: SelectProperty vs SelectPropertyValue

notion_client/lib/api_objects.py
```diff
@@ -0,0 +1,430 @@
...
+        return rollup
+
+
+@dataclass
+class NumberPropertyValue(PropertyValue):
+    number: float
+
+
+@dataclass
+class SelectPropertyValue(PropertyValue):
```
MrCredible, line 241: See previous comment about PropertyValue above.

This class is almost identical to SelectProperty and its use of SelectOption.

Spectrem12:
> > Another question, is it possible to make a pr to a pr? I would love to change small things here.
> 
> +1
> Great work @Jeadie !!
> I would love to do the same. It feels like breaking the work and the size of the PR is the way to remove the bottleneck. Please let me know how I can start contributing

Yes, If help is needed, I can support part of the effort as well.
This is a large amount of work, but it will be REALLY nice once it is complete.

Also, @ramnes , can we create a project for the writing to Notion portion?
So we can show the features that others may already have in progress..?

Jeadie:
I apologise, I have had no bandwidth for this until now. I believe I should be able to have this up considering your comments this week.

notion_client/lib/custom_enums.py
Jeadie: Great idea. Adding it in.

Jeadie:
to jheddings@ comment. In a similar way in that we deserialise the JSON response into pythonic classes, I am wanting to run a recursive like serialiser whereby every pythonic object can return it's JSON representation that will be correctly formatted for the API.

notion_client/lib/api_objects.py
Jeadie: I don't think the default use case makes it a good candidate for inheritance. Although they definitely share some logic that should be pulled out. 

I don't think it's a good candidate because when we receive page JSONs on an API request, we have no context to the additional keys needed to populate the Property Values (e.g. for a number we don't know the format of a Property Value that is received in an database API call). Thus we have alot of nulls in our Page objects and redundancy in keeping the state in both the database objects and page objects locally (for example, what happens if I change the format of a Property Value of a page; in fact how would that actually work in terms of updating the state across all PropertyValues with the one common Property).

aahnik:
> to jheddings@ comment. In a similar way in that we deserialise the JSON response into pythonic classes, I am wanting to run a recursive like serialiser whereby every pythonic object can return it's JSON representation that will be correctly formatted for the API. 

Why don't we use pydantic ?

Spectrem12:
Is this pull request still active?

ramnes:
I guess @Jeadie probably has other stuff to do at the moment; what's the real question? :)

Spectrem12:
I have been debating on trying to complete this PR but don't want to work on it if additional work has already been completed.

Trying to see if it's worth my time to look into

ramnes:
I'm going to close this one due to inactivity from OP and lack of bandwidth on my side, even thought I'll probably come back to it later. Check out @jheddings' library in the meantime. :)

PR #12, opened 2021-05-15
Use tox to test in multiple python versions locally

aahnik:
> At its core tox provides a convenient way to run arbitrary commands in isolated environments to serve as a single entry point for build, test, and release activities.

the GitHub actions matrix is cool for running tests in multiple python versions, in the ci.

But sometimes **developers may need to run tests on their local machine, especially while debugging**. Tox is the industry standard for running tests with multiple python versions.

tox_demo

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,3 +1,4 @@
 -e .
 pre-commit
 pytest
+tox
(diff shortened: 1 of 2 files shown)


aahnik (commented):
i have rechecked the changes to make sure that nothing breaks

ramnes:
Cool thanks!

PR #13, opened 2021-05-15
Fix the readme and add classifiers to setup.py

aahnik:
- added trove classifiers to setup.py https://pypi.org/classifiers
- add a badge to show supported python versions. Currently, this will show `missing`, but, once the changes to `setup.py` are pushed to PyPI, the badge will show something like this 
image
(for our case, 3.7 to 3.9)

- fix the emojis. the `:emoji:` is not supported in other platforms like PyPI.  (I have used the raw emojis, which are supported everywhere ✨)
image

(you may use vs code extension emoji sense for writing emojis)

- fix errors raised by [markdownlint](https://github.com/DavidAnson/markdownlint). mostly line length exceeded error.

These changes don't change the rendered README. but makes the raw readme, comply to markdownlint, and more human-readable. In some portions like the table and the place where HTML is used, markdownlint is disabled, as complying is impossible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+<!-- markdownlint-disable -->
 <div align="center">
     <h1>Notion SDK for Python</h1>
     <p>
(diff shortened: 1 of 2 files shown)


aahnik (commented):
I have re-checked the changes to make sure that nothing breaks.

ramnes:
Thanks, looking good overall! I didn't know about markdownlint, that's neat. But if we really want to use it, then it should be added to the pre-commit hooks, and thus CI. :)

aahnik:
Hi @ramnes, currently there is no way to add markdownlint alone to gh actions. But GitHub officially offers [superlint](https://github.com/github/super-linter), which includes black, isort, etc.

aahnik:
We may divide the ci into two parts:

`lint.yml` for code quality 

and `test.yml` for testing and uploading coverage results to codecov. (code coverage badge on readme)

ramnes:
superlint looks good, I didn't know about it either. :)

I'm good to add it in another PR if we can just disable pylint, which I have bad experiences with (i.e. you spend more time making it happy than the time it helps you win by giving you good insights). Let's merge this, then!

About the CI split, I'm not sure to understand, what's the benefit? I have the impression that it would only make it (much) slower with no real win.

aahnik:
workflow run in parallel in Github. So the code quality ci, and tests ci, will run parallel. We could know exactly which ci failed. code quality or tests. A pr that fails tests is dangerous, code quality can still be excused sometimes. Plus separate badges for code quality and tests. We don't want our tests to show failing when only quality fails.

ramnes:
Closing the PR, I merged it manually. Thanks again!

aahnik:
these three badges will be great

image

ramnes:
> workflow run in parallel in Github. So the code quality ci, and tests ci, will run parallel

Ah yeah, right!

> Plus separate badges for code quality and tests.

Alright, I'm in. :blush:

aahnik:
> Closing the PR, I merged it manually. Thanks again!

that's a neat technique. How to do this? Did you do this using the command line, or the GitHub UI ? can you provide any link to study this?

This technique will be useful in some of my other repos, where I want to edit the commits of contributors, maybe commit messages, as you did.

ramnes:
I just pulled your branch locally, ran an interactive rebase to change what I wanted to, and merged. :)

So something in these lines:
```
git checkout -B aahnik-patch-2
git pull git://github.com/aahnik/notion-sdk-py.git aahnik-patch-2
git rebase -i main
# moving commits around, fixup, reword
git checkout -
git merge aahnik-patch-2
git push
```

aahnik:
thank you so much for explaining.

PR #14, opened 2021-05-15
add errors (following notion-sdk-js's errors.ts)

Blue-Hope:
I'm not that familliar with typing with python so review on my typing objects will be appreciated very well


notion_client/client.py
```diff
@@ -61,8 +62,17 @@ def _build_request(self, method, path, body):
         return self.client.build_request(method, path, json=body)
 
     def request(self, path, method, query=None, body=None, auth=None):
-        request = self._build_request(method, path, body)
-        return self.client.send(request)
+        try:
+            request = self._build_request(method, path, body)
+            response = self.client.send(request)
+            response.raise_for_status()
+            return response
```
ramnes: Is it possible that we handle correctly an error that is raised by any other line than `response.raise_for_status()`? If not, we should rather have these other lines outside the try-except to let real errors (i.e. not related to the HTTP status code) being raised directly.

notion_client/client.py
```diff
@@ -84,7 +94,15 @@ def __init__(
...
+            async with self.client as client:
+                response = await client.send(request)
+            response.raise_for_status()
+            return response
+        except Exception as e:
+            request_error = build_request_error(e)
+            if request_error is None:
+                raise e
+
+            raise request_error
```
ramnes: Can we put the whole try-except logic in another method to avoid duplication between the clients? e.g. `Client._check_response(response)` or something like this.

ramnes (changes requested):
Thank you really much for the excellent work! Just two minor things I would like to see changed and I'm good to go!

notion_client/client.py
```diff
@@ -61,8 +58,17 @@ def _build_request(self, method, path, body):
...
+            request = self._build_request(method, path, body)
+            response = self.client.send(request)
+            response.raise_for_status()
+            return response
+        except Exception as e:
+            request_error = build_request_error(e)
+            if request_error is None:
+                raise e
+
+            raise request_error
```
ramnes: Would it be easier to read with something like this?

```suggestion
        except Exception as error:
            raise build_request_error(error) or error
```

blue-hope: sounds great

notion_client/client.py
blue-hope: I tested that If not `raise_for_status()`, HTTPStatusError is not occured and 400 / 500 errors will be ignored so we can't parse Notion's specified known errors. 
If not implemented, checking 400 / 500 errors logic must be specified on our code.
So rather than this way, I'd rather use httpx's parsing 400 / 500 logic.
400 / 500 errors that notion not specified will be filtered on here:
```py
def parse_api_error_response_body(body: dict) -> Union[APIErrorResponseBody, None]:
    if not is_api_error_code(body["code"]):
        return None
```
on errors.py line 94

notion_client/client.py
blue-hope: That sounds good

notion_client/client.py
ramnes: Yeah I'm good with `raise_for_status()`! I'm just saying we could have less lines of code in the try block. :)

notion_client/client.py
blue-hope: changed on latest commits

notion_client/client.py
blue-hope: changed on latest commits

notion_client/client.py
```diff
@@ -87,4 +96,6 @@ async def request(self, path, method, query=None, body=None, auth=None):
         request = self._build_request(method, path, query, body)
         async with self.client as client:
             response = await client.send(request)
+        response.raise_for_status()
```
ramnes: I guess you forgot that one!

notion_client/client.py
ramnes: Awesome, thanks!

notion_client/client.py
blue-hope: ah sorry

ramnes (approved):

ramnes:
Thanks again!

PR #16, opened 2021-05-15
add markdownlint to pre-commit

aahnik:
we don't need a GitHub superliner. That's too complex. I was wrong about the thing that markdownlint cant be individually added.


aahnik (commented):
I have experimented after adding this to pre-commit. everything works as expected.

ramnes:
Nice, thanks!

Issue #8, opened 2021-05-14
100% coverage


ramnes:
It's a noble goal, no matter what people say. :)

Blue-Hope:
#20 maybe helpful for goal 100% coverage :) how do you guys think?

Blue-Hope:
I create my new google id for this repo & create new notion accounts.
btw, we can't create databases / properties by official api. So I prepared pre-defined one and used it.
I'll request PR on this feature soon. Do you have idea for above problem?
(I think it's not matter for using pre-defined one. for 100% coverage, I prepared all properties for one database lol )

ramnes:
Hey @Blue-Hope, are you still willing to work on this? We started writing a few tests but there is still a lot to do!

Note to myself: release 1.0.0 when this is done.

Blue-Hope:
@ramnes Sure, I'd like to. Btw, can I refer to your written test?

ramnes:
What do you mean? You can of course write something similar to what has been written already, if that's the question. :)

PR #17, opened 2021-05-15
Measuring code coverage, and split ci workflow

aahnik:
We need to measure coverage if we want to walk the path of #8. And as discussed in https://github.com/ramnes/notion-sdk-py/pull/13#issuecomment-841625595, ci workflow split is implemented.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: CI
+name: Tests
 
 on:
   push:
@@ -18,10 +18,13 @@ jobs:
     steps:
 
     - uses: actions/checkout@v2
-    - name: Use Python ${{ matrix.node-version }}
+    - name: Use Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
     - run: pip install -r requirements-dev.txt
-    - run: pre-commit run --all-files
-    - run: pytest
+    - run: pytest --cov=./notion-client --cov-report=xml
+    - uses: codecov/codecov-action@v1
+      with:
+        env_vars: PYTHON
+        fail_ci_if_error: true
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - run: pip install -r requirements-dev.txt
-    - run: pytest --cov=./notion-client --cov-report=xml
+    - run: pytest --cov=./notion_client --cov-report=xml
     - uses: codecov/codecov-action@v1
       with:
         env_vars: PYTHON
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - run: pip install -r requirements-dev.txt
-    - run: pytest --cov=./notion_client --cov-report=xml
+    - run: pytest
     - uses: codecov/codecov-action@v1
       with:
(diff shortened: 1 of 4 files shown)


aahnik:
uploading to codecov failed. hi @ramnes, do you have a codecov account?

```
{'detail': ErrorDetail(string='Could not find a repository, try using upload token', code='not_found')}
```

If you create a codecov account, I think this will be fixed. Because as per codecov docs, public repos don't require an upload token.

aahnik:
this https://app.codecov.io/gh/ramnes/notion-sdk-py shows empty. If you create a codecov, account, I think you will be able to figure out the reason.

In a separate public repo where I tested, codecov did not require an upload token.

.github/workflows/test.yml
```diff
@@ -18,10 +18,13 @@ jobs:
...
     - uses: actions/checkout@v2
-    - name: Use Python ${{ matrix.node-version }}
+    - name: Use Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
     - run: pip install -r requirements-dev.txt
-    - run: pre-commit run --all-files
-    - run: pytest
+    - run: pytest --cov=./notion_client --cov-report=xml
```
ramnes: Let's configure pytest in setup.cfg so that it's not needed to write this! And why the `--cov-report=xml` rather than `term-missing`?

requirements-dev.txt
```diff
@@ -2,3 +2,5 @@
 pre-commit
 pytest
 tox
+coverage
+pytest-cov
```
ramnes, line 5: Let's keep requirements in alphabetical order. :)

ramnes (commented):
LGTM overall, thanks!

.github/workflows/test.yml
aahnik: no specific reason. I saw xml to be used in example given by codecov. I think codecov supports only XML. Not sure.

As per the docs of the pytest-cov plugin,

```text
--cov-report=TYPE     Type of report to generate: term, term-missing, annotate, html, xml (multi-allowed). term,
                        term-missing may be followed by ":skip-covered". annotate, html and xml may be followed by
                        ":DEST" where DEST specifies the output location. Use --cov-report= to not generate any
                        output.
```

multiple types are allowed. So what about, using both XML and term-missing.

aahnik: > Let's configure pytest in setup.cfg so that it's not needed to write this!

that's a great idea

aahnik: but here is a warning given by pytest

image

https://docs.pytest.org/en/6.2.x/customize.html#setup-cfg

aahnik:
hi @ramnes, how did you fix this https://github.com/ramnes/notion-sdk-py/pull/17#issuecomment-841633527, by creating a codecov account?

ramnes:
Yeah, and adding the repository.

.github/workflows/test.yml
ramnes: The use case is simple enough, let's avoid a pyproject.toml for now if we can. We'll move everything in a pyproject.toml when we'll switch to Poetry or another PEP 517 manager.

ramnes: > multiple types are allowed. So what about, using both XML and term-missing.

Let's try without XML first, and if Codecov really needs it, put it back. :)

aahnik:
Hi @ramnes, I have used both xml and term-missing.

[image](
image
)

aahnik (commented):
I tested all the changes, to ensure nothing broke. This is ready to be merged.

ramnes:
Thanks again!

PR #20, opened 2021-05-15
Add basic test with notion api mock

Blue-Hope:
for 100% coverage, I added some mocks on Notion API


aahnik:
instead of adding conftest.py to the root of the project, why not use a `tests` directory?

tests/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+            status_code = StatusCode.InternalServerError.value
+        elif code == APIErrorCode.ServiceUnavailable:
+            status_code = StatusCode.ServiceUnavailable.value
+        else:
+            status_code = StatusCode.Ok.value
+        super().__init__(
+            request=request,
+            status_code=status_code,
+            json=json or {"code": code.value, "message": ""},
+        )
```
ramnes, line 54: This could be made simpler with something like: 

```python
status_codes = {
    APIErrorCode.InvalidJSON: 400,
    APIErrorCode.InvalidRequestURL: 400
    ...
}

...
        super().__init__(
            request=request,
            status_code=status_codes.get(code, 200),
            json=json or {"code": code.value, "message": ""},
        )
```

ramnes: (that would remove the whole if tree)

ramnes:
I think maintaining a proper mock of all Notion endpoints will be super tedious, especially since they're a private company that probably won't communicate much when things in the API change.

What would you think of setting up a test account and using [VCR.py](https://github.com/kevin1024/vcrpy)? I've never used it yet but it does seem like the perfect use case to me to try it!

Blue-Hope:
really nice idea 😀 I'll close this PR and make new one with VCR soon 👋

aahnik:
hi @ramnes , vcrpy seems not to support httpx 

[image](https://vcrpy.readthedocs.io/en/latest/installation.html)

[Ref](https://vcrpy.readthedocs.io/en/latest/installation.html)

[Open issues](https://github.com/kevin1024/vcrpy/issues?q=is%3Aissue+httpx+is%3Aopen)

ramnes:
Looks like it works in most cases, see the last commits. :)

PR #21, opened 2021-05-15
Leverage type hints

### LINKED ISSUE #6: Leverage type hints
- [ ] Add complete type hints for all signatures
- [ ] Bundle a `py.typed` file
- [ ] Make Mypy run in the CI

aahnik:
fixes #6 

- [x] Make Mypy run in the CI
- [x]  Bundle a `py.typed` file
- [x] Add type hints wherever possible 
- [ ] satisfy mypy fully

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,4 +16,6 @@ repos:
   rev: v0.812
   hooks:
   - id: mypy
+    additional_dependencies: ["httpx"]
+    args: []
     files: "^notion_client\/.*"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,3 +12,8 @@ repos:
   rev: v0.27.1
   hooks:
   - id: markdownlint
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: v0.812
+  hooks:
+  - id: mypy
+      
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,4 +16,4 @@ repos:
   rev: v0.812
   hooks:
   - id: mypy
-      
+    files: "^notion_client\/.*"
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Configuration for code editors
+.vscode
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/notion_client/__init__.py
+++ b/notion_client/__init__.py
@@ -1,2 +1,4 @@
 from .client import AsyncClient, Client
 from .errors import APIErrorCode, APIResponseError
+
+__all__ = ["AsyncClient", "Client", "APIErrorCode", "APIResponseError"]
--- a/notion_client/api_endpoints.py
+++ b/notion_client/api_endpoints.py
@@ -9,7 +9,7 @@ if TYPE_CHECKING:
 
 
 class Endpoint:
-    def __init__(self, parent: Union["BaseClient", "Client", "AsyncClient"]) -> None:
+    def __init__(self, parent: "BaseClient") -> None:
         self.parent = parent
 
 
--- a/notion_client/api_endpoints.py
+++ b/notion_client/api_endpoints.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Any, Coroutine, Union
+from typing import TYPE_CHECKING, Any, Awaitable, Union
 
 from httpx import Response
 
@@ -16,7 +16,7 @@ class Endpoint:
 class BlocksChildrenEndpoint(Endpoint):
     def append(
         self, block_id: str, **kwargs: Any
-    ) -> Union[Response, Coroutine[Any, Any, Union[Response, Any]]]:
+    ) -> Union[Response, Awaitable[Response]]:
         return self.parent.request(
             path=f"blocks/{block_id}/children",
             method="PATCH",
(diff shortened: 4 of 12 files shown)


ramnes:
Have a look at [this comment on Reddit](https://www.reddit.com/r/Notion/comments/ncjy7v/let_me_introduce_you_to_notionsdkpy_a_python/gy5rmvd?utm_source=share&utm_medium=web2x&context=3). The first point they make about the current classes inheritance might bite you.

mypy.ini
```diff
@@ -0,0 +1,17 @@
...
+disallow_incomplete_defs = True
+check_untyped_defs = True
+disallow_untyped_decorators = True
+no_implicit_optional = True
+warn_redundant_casts = True
+warn_unused_ignores = True
+warn_return_any = True
+implicit_reexport = False
+strict_equality = True
+# --strict end
```
ramnes: Can't we throw this in setup.cfg?

.pre-commit-config.yaml
```diff
@@ -12,3 +12,8 @@ repos:
   rev: v0.27.1
   hooks:
   - id: markdownlint
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: v0.812
+  hooks:
+    - id: mypy
```
ramnes: Inconsistent indentation

aahnik: extremely sorry for overlooking that.

mypy.ini
aahnik: that's a good idea. I always love keeping the root directory clean and minimal.

aahnik:
> I see AsyncClient inherits from Client and overrides its request method. Having recently worked on building a sync/async client, I can tell you that while this works, you'll never be able to make mypy happy, and it kind of breaks the Liskov substitution principle.

> What I'd recommend instead is that there's a BaseClient abstract class which concerns itself with all non-transport logic, and that Client and AsyncClient then subclass the base client (but not each other) and implement their transport methods, using the base class' methods for validating/parsing data, etc.

this idea by @pedrovhb is great. I think we should do that.

this would fix this

> notion_client/client.py:93: error: Argument 2 to "__init__" of "Client" has incompatible type "AsyncClient"; expected "Client"

the base client should accept both AsyncClient, as well as Client

aahnik:
hi @ramnes, I am a little less confident about ff9e260.

please review to make sure I have made no mistakes.

As explicit is better than implicit, I have added `-> None` to functions that return None.

ramnes:
LGTM so far overall but there are still a lot of Mypy errors to fix!

> structural changes need to made in the code. I am afraid of doing that, as it might lead to conflicts with other open PRs.

Don't be afraid, one of the two PR is only working in `examples/`, and the other one is a big WIP that will have to live with rebases anyway, and I'll give some help if needed. :)

Do you mind if I add commits to this branch to give a hand?

aahnik:
> Do you mind if I add commits to this branch to give a hand?

That will be great! Please do so.

aahnik:
All `mypy` issues fixed, except a few in `client.py` and `errors.py`

-----------

I made a `BaseClient` and restructured the code, to tackle a lot of `mypy` errors. Still some remains.


```shell
# client.py

notion_client/client.py:99: error: Argument 1 to "_check_response" of "Client" has incompatible type "Union[Response, Coroutine[Any, Any, Response]]"; expected "Response"
notion_client/client.py:103: error: Incompatible return value type (got "Union[Response, Coroutine[Any, Any, Response]]", expected "Response")
notion_client/client.py:123: error: "Coroutine[Any, Any, Response]" has no attribute "raise_for_status"
notion_client/client.py:140: error: Argument 1 to "_check_response" of "AsyncClient" has incompatible type "Response"; expected "Coroutine[Any, Any, Response]"
Found 4 errors in 1 file (checked 1 source file)

# trying to solve this issues, leads to other issues
```

I think the only way to tackle them, is to completely separate the Client and AsyncClient. But in that way, there will be a lot of code repetition. (already there is some code repeated. I am trying to make it more elegant, without changing the outside interface to Client and AsyncClient).

The problem is that for the AsncClient the functions return the type coroutine. I think an intelligent refactor is needed. Please help me out @ramnes 

-----------

The errors.py is still raising some problems.

```shell
# errors.py

notion_client/errors.py:89: error: "Exception" has no attribute "response"
notion_client/errors.py:91: error: "Exception" has no attribute "response"
notion_client/errors.py:92: error: "Exception" has no attribute "response"
Found 3 errors in 1 file (checked 1 source file)

```

aahnik:
Thanks for the commits. I was doing few things wrong, as I was driven by the urge to suppress mypy. I had a few unpushed commits, but I deleted them,as they were dangerous.(huge code repition)

I have no idea how to fix, the remaining mypy errors. 😢 Its all up to you now.

ramnes:
Yeah I can take over, no worries, thanks a lot for the work so far!

aahnik:
How did mypy pass in CI ?

Running mypy locally still gives two errors:


```shell
➜ mypy notion_client
notion_client/client.py:128: error: Argument 1 to "_check_response" of "BaseClient" has incompatible type "Union[Response, Coroutine[Any, Any, Response]]"; expected "Response"
notion_client/client.py:129: error: Incompatible return value type (got "Union[Response, Coroutine[Any, Any, Response]]", expected "Response")
Found 2 errors in 1 file (checked 6 source files)
```

ramnes:
Yeah, I wondered that as well. Answer [here](https://github.com/pre-commit/pre-commit/issues/1580#issuecomment-679413936). :)

ramnes:
Let's go! :tada:

Thank you again for your very motivating efforts @aahnik.

PR #24, opened 2021-05-15
Discord-Notion Integration Example

FruitVodka:
This is a Discord bot to port book recommendations from a text channel on Discord to a Notion database. As an intermediate step, it looks up the book title using the Google Books API and collects some details to add to the database.

We tried using multi-select to set "Genres" for the books but since the Notion API doesn't support auto-addition of options for select and multi-select, had to scrap that idea for now! :(

The Discord bot in action!
SS1
Here's the DB template: [https://www.notion.so/e2b278fd05574df694833e6790e02340?v=34c458559056411d8cbd3bc3a7f406d6](https://www.notion.so/e2b278fd05574df694833e6790e02340?v=34c458559056411d8cbd3bc3a7f406d6)
You can see "Master of the Game" has been added to the DB!!


aahnik (changes requested):
`.env` is conventionally used to store secret environment variables. This should not be committed to source control. 

Revert `.gitignore` to the original state. Delete the `.env` file.

Add an example of `.env`  inside the README.md

aahnik:
@FruitVodka  your work is wonderful. but it is a full-fledged project. it will be best if you create a separate repo(by that you have more control, and can push more often). interesting examples like this could be featured in the readme of this project.


examples directory is a good idea. but examples living inside this repo should be minimal, and use this library only. examples inside this repo should be mainly for learning purposes, and to show the full use of this library(which will actively change over this month)

@ramnes, what do you think?

FruitVodka:
@aahnik Thank you for your feedback! I do agree this is a bit more complex than an ideal example should be 🤔 
@ramnes do you agree? Would you be willing to feature this in the readme of notion-sdk-py?

ramnes:
Hey there, sorry for the long time to answer, I wasn't sure what to answer. Because yeah, I agree with both of you, it's quite big for an example, and it's quite cool at the same time!

With the combination of 1. a dependency on other libraries we don't have control on, 2. a niche use-case, 3. a rather large code size (for an example) and 4. no test, the code will be hard to maintain correctly over time. If you can kill at least two of these four points, I would probably consider merging it.

But if you don't want to take more time on this and are tempted to make it your own repository, I'd be happy to add a link in the README too!

aahnik:
Hi @ramnes , I think we should leave idea of making it close to the js SDK. Rather priority should be to make it more pythonic, and follow the footsteps of major api wrappers written in python.

ag0n1k:
Hi everyone!

If it counts, I bet to add some opinion here:

I'm the person, who
1. Wanted to automate routine stuff at notion
2. Tired of waiting for the official api
3. Found and create application with notion-py (https://github.com/jamalex/notion-py)
4. Found this repo without samples, but with great activity and started waiting for its "stable" version =)

The next task of my project is to try to move to this sdk, and this example is huge and great, but after I read it, it's not the answer that I was looking for. It's great example for "already created projects with sdk" or for people who integrate notion with discord, but not for the sdk examples.

aahnik:
> a dependency on other libraries we don't have control on

this is unavoidable if we want to integrate notion with anything else. Notion js sdk examples use external dependencies (like the GitHub issue sync uses octokit). If we don't allow external deps, then, all examples will be dealing with Notion alone, and not with any other product. That's a good idea.

- examples inside this repo must use the latest stable version of notion-sdk-py
- usage outside this repo can use whatever version the author likes

See telethon(python client for telegram), they have a [wiki page](https://github.com/LonamiWebs/Telethon/wiki/Projects-using-Telethon) where all examples are listed. That's a good idea, I think. 

💡 **Idea**

One day there will be 1000s of projects using this library. A cool idea would be to maintain a notion database of all GitHub projects that used the tag "notion-sdk-py". 

We could run a scheduled GitHub action every day, and update the list of projects using this lib.  

The top 5 most starred projects of this database can be featured in the readme. And a link to the database could be given as "more projects ...". The readme updation can be done automatically, using gh actions.

ramnes:
I really like Telethon's setup, with a page in the docs referencing other projects, and an example directory with simple ones. Let's do it that way! Is it good for you @FruitVodka?

ramnes:
> We could run a scheduled GitHub action every day, and update the list of projects using this lib.

I think we should rather keep a curated list of things we like. GitHub has our back for a complete list with the "Used by" insight.

aahnik:
> GitHub has our back for a complete list with the "Used by" insight.

There is no guarantee that a project will always appear in "Used by". GitHub currently does not support pyproject.toml, or poetry.lock,

https://github.com/python-poetry/poetry/discussions/3986

FruitVodka:
> I really like Telethon's setup, with a page in the docs referencing other projects, and an example directory with simple ones. Let's do it that way! Is it good for you @FruitVodka?

Yeah, this seems good! I'll open a separate repo for this code and let you know.
I can also work on a better-suited example, I still think an example integration might be the way to go, but maybe we can do a much simpler one. What are your views?

FruitVodka:
Hi @ramnes here's the link to my repo: [https://github.com/FruitVodka/notion-discord-book-recommendations](https://github.com/FruitVodka/notion-discord-book-recommendations).
I've added it to the wiki page as in the Telethon example. I also added a small table of contents for easier navigations through the wiki page. Do let me know if that works!

ramnes:
Yeah perfect, thanks!

PR #28, opened 2021-05-17
Fix variable name mistake in README.md

aahnik:
timeoutMs -> timeout_ms

the code is already correct

https://github.com/ramnes/notion-sdk-py/blob/1e676511db7a0c6627d5cf8c3ba7b1940118e043/notion_client/client.py#L21


ramnes:
Woops. :)

PR #35, opened 2021-05-20
Add status badges to readme for code quality and tests

aahnik:
I could not do for codecov badge , you need to do that yourself @ramnes, as only admins can copy the URL (which contains a unique token)


ramnes:
Thanks!

PR #40, opened 2021-05-25
Add url helpers

aahnik:
It's a very common requirement to access a page by its URL.  We could definitely do a search, but multiple pages/databases may exist with the same name.

When a python application reads/writes to a Notion page or database, fetching the exact object is required. The user could easily share the URL with the application, and the app can determine the id from the URL. With the id, the developer can retrieve the exact object.

- I have added helper functions to get the id from the URL or get the URL from the id.
- I have added `tests` for these functions. As nothing happens over the network, we don't need [vcrpy](https://github.com/kevin1024/vcrpy) for this
- I have added `helpers` to the reference section of the docs site.

Usage of this new feature

```python
from notion_client.helpers import get_id

my_url = "..."

my_id = get_id(my_url)

```

This is inspired by discussion #31. Special thanks to @Blucknote  for this answer https://github.com/ramnes/notion-sdk-py/discussions/31#discussioncomment-761925

Relevant section of Notion Docs 
- [Database id](https://developers.notion.com/docs/working-with-databases#adding-pages-to-a-database)
- [Page id](https://developers.notion.com/docs/working-with-page-content#creating-a-page-with-content)


notion_client/helpers.py
```diff
@@ -6,3 +6,14 @@
...
+
+def get_url(obj_id: str) -> str:
+    """Return url for the object of given id."""
+    return "https://www.notion.so/" + "".join(obj_id.split("-"))
+
+
+def get_id(url: str) -> str:
+    """Return the id of the object of given url."""
+    slug: str = url[-32:]
+    return "-".join([slug[0:8], slug[8:12], slug[12:16], slug[16:20], slug[20:32]])
```
ramnes: This is really flimsy and based on a lot of expectations that could be wrong. Try to see the output of something like `notion.so/123`. It should not return something broken but rather None, or raise an error. You probably want to use `urllib.parse` here. :)

tests/test_helpers.py
```diff
@@ -0,0 +1,22 @@
+from notion_client.helpers import get_id, get_url, pick
+
+
+def test_pick():
+    my_dict = {"Product": "Notion", "API": 2021, "python-sdk": "ramnes"}
+    assert pick(my_dict, *["Product"]) == {"Product": "Notion"}
+    assert pick(my_dict, *["API", "python-sdk"]) == {
+        "API": 2021,
+        "python-sdk": "ramnes",
+    }
```
ramnes: I don't see a reason to build a list and unpack it at the same time here:

```suggestion
    assert pick(my_dict, "Product") == {"Product": "Notion"}
    assert pick(my_dict, "API", "python-sdk") == {
        "API": 2021,
        "python-sdk": "ramnes",
    }
```

ramnes (changes requested):

notion_client/helpers.py
aahnik: Sure, I will do so. I will also add test cases, to check if the function is raising proper errors when a wrong URL is given.

aahnik (commented):
I have used `urlib.parse` It's fantastic, as it does all URL-related validation.

and `UUID` does all validation regarding the 32 characters long hex string.

the dash and non-dash format conversion is also handled by that.

aahnik (commented):
i re-checked all changes to make sure nothing breaks!

ramnes:
Thanks!

PR #54, opened 2021-06-08
Update the usage of codecov action in test.yml

aahnik:
-> use codecov token (you must set the token in github secrets)
-> dont fail ci on codecov error

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,5 +26,6 @@ jobs:
     - run: pytest
     - uses: codecov/codecov-action@v1
       with:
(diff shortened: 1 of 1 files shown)


ramnes:
https://github.com/ramnes/notion-sdk-py/pull/51#issuecomment-857133586

Thanks for the investigation though!

aahnik:
Codecov guys have not fixed it yet. The issue is still open https://github.com/codecov/codecov-action/issues/330

It is unreliable. Sometimes works sometimes does not. Using token will ensure no failure.

PR #60, opened 2021-06-20
Remove httpx maximum version check

nlecoy:
Hey @ramnes, thank you for this great work.
I'm opening this PR to remove the maximum version of `httpx` so we don't have conflicts with projects using your library & `httpx`.

What do you think?
Have a nice day


ramnes:
Just took a look at the changelog and it shouldn't be a problem, yes. Thanks!

Issue #65, opened 2021-07-28
Update method of PagesEndpoint class never handles archived pages


Thompsonmina:
Calling something like 
`
	archivedpage = notion.pages.update(page_id, **{"archived":True, "properties":properties})
`
will never result in the page being archived. Tested calling the same route on curl and it works

Did some digging and i think i figured out where the bug is from,
```
def update(self, page_id: str, **kwargs: Any) -> SyncAsync[Any]:
        return self.parent.request(
            path=f"pages/{page_id}",
            method="PATCH",
            body=pick(kwargs, "properties"),
            auth=kwargs.get("auth"),
        )
```
The line `body=pick(kwargs, "properties")` in the update method doesn't pick up the archived keyword from kwargs, resulting in only the properties item being added to body

ramnes:
Fixed in with #66

PR #66, opened 2021-07-28
Fix very minor bug on Update method of api_enpoints.PagesEndpoint

Thompsonmina:
referencing this issue #65
took the liberty of adding the archived arg s


ramnes:
Thanks!

Issue #78, opened 2021-09-06
Cannot update icon and cover


fedor-ivn:
### The problem
In the new version, Notion added an icon and image cover support and consequently fields in API, but the SDK doesn't support this feature (and fields) yet.

### Possible solution
I suppose the problem is here in line 147:
https://github.com/ramnes/notion-sdk-py/blob/4dc60c491b51d152fdb93caa0efd00d91f30adb9/notion_client/api_endpoints.py#L135-L149

I guess It should be changed like that:
```py
# "icon" and "cover" fields were added
body=pick(kwargs, "parent", "properties", "children", "icon", "cover"),
```

Dr0p42:
👋 looking forward to have this in a release 🎉

Thank you @fedor-ivn @al1p 👏

PR #85, opened 2021-10-09
fix: create  & update pages and databases endpoint for icon & cover

al1p:
As suggested by @fedor-ivn
https://github.com/ramnes/notion-sdk-py/issues/78


ramnes:
Why closing? :)

Issue #82, opened 2021-10-07
Missing examples of creating databases


stereorobin:
Hello,
I'm having problems creating a database, mostly due to lack of any examples.
I can create a page just fine, but am having problems with databases.
I think I just get type/format of the title and properties wrong.

Would be great to have an example of how to create a database with a title and a text property.

Thanks for your great work!

ramnes:
Uh, this has been fixed a long time ago actually! Thanks for raising the issue.

PR #89, opened 2021-10-14
Example to create database

al1p:
as requested by @stereorobin in #82


ramnes:
How is it related to what the example does so far? I feel like this should be a completely new example with its own directory.

PR #93, opened 2021-11-12
Updates Client Options in Readme

nyoungstudios:
- updates client options in readme and reorders the table to the same order as the [code](https://github.com/ramnes/notion-sdk-py/blob/11e684c8b5dc10ec6791c23ad56b8f3943142786/notion_client/client.py#L45-L50)


ramnes:
Hi @nyoungstudios! 👋 

Thank you for the PR, but I prefer to keep the table as is. I don't particularly want people to play with the `notion_version` argument.

Issue #61, opened 2021-06-23
TimeoutException should handled during `send()` not in `_parse_reponse`


waynerv:
```
def _parse_response(self, response: Response) -> Any:
        try:
            response.raise_for_status()
        except httpx.TimeoutException as error:
            if is_timeout_error(error):
                raise RequestTimeoutError()
            raise
        except httpx.HTTPStatusError as error:
            body = error.response.json()
            code = body.get("code")
            if is_api_error_code(code):
                body = APIErrorResponseBody(code=code, message=body["message"])
                raise APIResponseError(error.response, body)
            raise HTTPResponseError(error.response)

        return response.json()
```
Timeout exception handling logic above in `BaseClient._parse_response()` would never run, becuase `raise_for_status()` won't raise `TimeoutException`.

mutt0-ds:
@ramnes I must warn you that I'm just a beginner, but I love the work you are doing and I'm trying to help whenever is possible. Please take my code with a grain of salt ahah

waynerv:
@mutt0-ds You should make a pull request, this commit is still in your forked repo. 

Just remove this codes is not enough, Timeout Error should be properly handled.

mutt0-ds:
Thanks for the feedback, I'll try to work on it in my testing environment and then make a pull request

ramnes:
Hey @mutt0-ds, are you still willing to work on this one? :)

mutt0-ds:
Hey @ramnes sorry, I'll try to work on it next week since I've been using the library for months and I know it better now!

ramnes:
Fixed in #61

PR #97, opened 2021-12-15
Issue #61 - Handled Timeout Errors

mutt0-ds:
As discussed in the issue #61, since response.raise_for_status() is never raising a TimeoutException, I created a function for checking if the status code belongs in the "Timeout" category.

I'm curious to hear your feedbacks, I hope that my approach makes sense :)

### DIFF
--- a/notion_client/client.py
+++ b/notion_client/client.py
@@ -20,6 +20,7 @@ from notion_client.errors import (
     HTTPResponseError,
     RequestTimeoutError,
     is_api_error_code,
+    is_timeout_error_code,
 )
 from notion_client.logging import make_console_logger
 from notion_client.typing import SyncAsync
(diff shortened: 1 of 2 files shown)


ramnes:
Hey @mutt0-ds, thanks for the work!

The implementation looks wrong to me; isn't httpx raising an exception when there's a timeout? If so, we need to catch it when doing the request, not trying to look at the HTTP code, that we'll only get if we actually get an answer from the remote machine.

mutt0-ds:
Hey @ramnes thank you for your reply! 
You are right, I need to handle the timeout not in the raise_for_status() function, but first, in when it's sending the request!
So I handled the TimeoutException in request(), since the TimeoutExpection is raised there.  
Let me know if this time I understood correctly!

ramnes:
Yeah, that's it! The asynchronous client also needs it though.

mutt0-ds:
Updater for the AsyncClient too :)

ramnes:
Great, thanks!

Issue #94, opened 2021-11-29
in query empty start_cursor option throws error


rzylius:
hi,

I try to make a query:

```
next_cursor = None
results = notion.databases.query(
        **{
            "database_id": DB_GMB_ID,
            "start_cursor": next_cursor
        }
    )
```
I receive the error:

```
APIResponseError: body failed validation: body.start_cursor should be a string or `undefined`, instead was `null`.
```

"" produces the same result.

What value should be used to satisfy undefined start_cursor?

ramnes:
Hello @rzylius, that's quite a funny one and I've no idea what Notion is expecting from us. I'll investigate eventually, please tell me if you find anything new in the meantime!

ramnes:
Removing that `pick` thing and using proper arguments would probably help.

nicobako:
I don't think this is an issue with notion_client. In the notion API documentation it states that `start_cursor` must be a string.

From https://developers.notion.com/reference/post-database-query

>   start_cursor
>   string
>   When supplied, returns a page of results starting after the cursor provided. If not supplied, this endpoint will return the first page of results.

It seems that it must either be specified as a string, or not  specified at all. `null` (or `None`, in Python), is not a valid value of `start_cursor`.

rzylius:
> It seems that start_cursor must either be specified as a string, or not specified at all. null (or None, in Python), is not a valid > value of start_cursor.

The error thrown "APIResponseError: body failed validation: body.start_cursor should be a string or `undefined`" says, that the value can be 'undefined'. This makes sense in javascript, but how 'undefined' works in Python?

nicobako:
> The error thrown "APIResponseError: body failed validation: body.start_cursor should be a string or `undefined`" says, that the value can be 'undefined'. This makes sense in javascript, but how 'undefined' works in Python?

There is no `undefined` in Python. Python has `None`, which is equivalent to `null`.

In Python, we have the concept of an *optional value*, where something can either be a *value* or `None`. For example, `x = 1` or `x = None`.  But, in both cases, a variable named `x` exists (i.e. `x` is not `undefined`).

nicobako:
I see two possible fixes here:

1. Remove usage `start_cursor=None` from your code
2. Modify `pick(...)` so that it will remove `start_cursor=None` for you

I'm not sure which one of these is the best solution... Perhaps others have an opinion.

rzylius:
[...]

PR #101, opened 2022-01-09
modified pick helper function to only keep keys if value is not None

nicobako:
@ramnes ,

I believe this should fix the issues mentioned here: #94

I haven't fully tested this yet, but I hope the GitHub Actions would be able to test this more fully.

Note, this is my first time contributing to this project, and I would love to contribute more. If there is anything else you would like me to do differently please let me know.

### DIFF
--- a/notion_client/api_endpoints.py
+++ b/notion_client/api_endpoints.py
@@ -141,7 +141,7 @@ class DatabasesEndpoint(Endpoint):
         return self.parent.request(
             path=f"databases/{database_id}/query",
             method="POST",
-            body=pick(kwargs, "filter", "sorts", "start_cursor", "page_size", keys_to_pop_if_none=["start_cursor"]),
+            body=pick(kwargs, "filter", "sorts", "start_cursor", "page_size"),
             auth=kwargs.get("auth"),
         )
 
--- a/notion_client/api_endpoints.py
+++ b/notion_client/api_endpoints.py
@@ -141,7 +141,7 @@ class DatabasesEndpoint(Endpoint):
         return self.parent.request(
             path=f"databases/{database_id}/query",
             method="POST",
-            body=pick(kwargs, "filter", "sorts", "start_cursor", "page_size"),
+            body=pick(kwargs, "filter", "sorts", "start_cursor", "page_size", keys_to_pop_if_none=["start_cursor"]),
             auth=kwargs.get("auth"),
         )
 
--- a/notion_client/helpers.py
+++ b/notion_client/helpers.py
@@ -1,12 +1,18 @@
 """Utility functions for notion-sdk-py."""
-from typing import Any, Dict
+from typing import Any, Dict, List
 from urllib.parse import urlparse
 from uuid import UUID
 
 
-def pick(base: Dict[Any, Any], *keys: str) -> Dict[Any, Any]:
+def pick(base: Dict[Any, Any], *keys: str, keys_to_pop_if_none: List[str]=[]) -> Dict[Any, Any]:
     """Return a dict composed of key value pairs for keys passed as args."""
-    return {key: base[key] for key in keys if key in base and base[key] is not None}
+    body = {key: base[key] for key in keys if key in base}
+
+    for key in keys_to_pop_if_none:
+        if key in body and body[key] is None:
+            body.pop(key)
+
+    return body
 
 
 def get_url(object_id: str) -> str:
--- a/notion_client/helpers.py
+++ b/notion_client/helpers.py
@@ -6,7 +6,7 @@ from uuid import UUID
 
 def pick(base: Dict[Any, Any], *keys: str) -> Dict[Any, Any]:
     """Return a dict composed of key value pairs for keys passed as args."""
-    return {key: base[key] for key in keys if key in base}
+    {key: base[key] for key in keys if key in base and base[key] is not None}
 
 
 def get_url(object_id: str) -> str:
--- a/notion_client/helpers.py
+++ b/notion_client/helpers.py
@@ -6,7 +6,7 @@ from uuid import UUID
 
 def pick(base: Dict[Any, Any], *keys: str) -> Dict[Any, Any]:
     """Return a dict composed of key value pairs for keys passed as args."""
-    {key: base[key] for key in keys if key in base and base[key] is not None}
+    return {key: base[key] for key in keys if key in base and base[key] is not None}
 
 
 def get_url(object_id: str) -> str:
--- a/notion_client/helpers.py
+++ b/notion_client/helpers.py
@@ -1,19 +1,12 @@
 """Utility functions for notion-sdk-py."""
-from typing import Any, Dict, List
+from typing import Any, Dict
 from urllib.parse import urlparse
 from uuid import UUID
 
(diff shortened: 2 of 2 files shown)


ramnes:
Hey there, thanks for opening that PR!

Yes, that would work in that scenario, but what if we really want to pass `None` as a value to Notion's API? Are we sure it can't happen?

nicobako:
Your absolutely right, it could happen that `None` is a valid value, in which case we would need to keep it.

From what I understand, in JavaScript, a variable can either

1. Have a value (aka be defined)
2. Be `null`
3. Be `undefined`

And each one of these has its own meaning in JavaScript. In python we can only handle defined and null variables, but not undefined variables.

In the notion API, I've seen cases where a variable can be either

* Defined or null
* Defined or undefined

But I have yet to see a situation where a variable can be either defined, null, or undefined...

What do you think about adding a little logic to the pick function so that, in cases where a variable must either be defined or undefined, the pick function would just remove it if it's `None`? Otherwise, it would just leave it alone.

nicobako:
@ramnes , I realize my description above might be difficult to follow, so I went ahead and implemented what I was talking about so that you can see it for yourself.

With these changes, key-value pairs with a value of `None` are *popped* by default. Key-value pairs that have a value of `None` are only kept if they are explicitly allowed via the `keys_to_keep_if_none`. I think this would work well with the Notion API, and would help solve the issues users have been experiencing.

Anyway, you may have better ideas. I'm open to suggestions.

nicobako:
Haha, I agree with you, sort of... Function arguments would be nice, but that still doesn't solve the real issue.

The real issue is that users want to specify, for example, `start_cursor=None` because this feels *Pythonic*. But, it's invalid to pass `start_cursor=null` to the Notion API. The [*workaround* you proposed](https://github.com/ramnes/notion-sdk-py/issues/94#issuecomment-993413583) is not *Pythonic* ... but it's a good workaround for now.

This pull request is really about improving the user experience while using `notion_client`. It's really about allowing users to specify: `variable=None`, and then silently removing `variable=None` before a request is made to the Notion API.
In fact, that's exactly what @thamshirtz recommended in [his issue comment](https://github.com/ramnes/notion-sdk-py/issues/94#issuecomment-1008271600).

> I think it is needed to remove the argument start_cursor internally if None is passed, because I don't see any situation where it is wanted to get this error response

nicobako:
Here's another idea of how we could fix this: introduce the concept of an `Undefined` object.

The Notion API relies heavily on `undefined` variables, but there is simply no such thing in Python. If we created a `notion_client.Undefined` object, then users could use that instead of `None`.

Then the `pick` function could easily remove any *key-value* pair where the value is `Undefined`, but leave anything with a value of `None`. What do you think?

nicobako:
> Did you really see a use case for passing `None` or was it just theory?

Actually, I think I completely made it up... sorry. I haven't seen a use case for passing `None`. 

The Notion API seems to rely heavily on either a variable being _defined_ or _undefined_, but not on it being _null_.

That's one reason why I think we can just have the `pick` function silently remove any *key-value* pairs with a value of `None`. I think this is a simple solution that will work very well, as far as we can tell. If we need to support some _key-value_ pair with a value of `None`, then we can either worry about that later, or use the `keys_to_keep_if_none`.

ramnes:
Let's revert to the state where you just ignored `None` values then. We'll update to something else if anyone complains.

nicobako:
@ramnes , just reverted everything back, as you said.

ramnes:
Merged, thanks!

ramnes:
Seems like we found a use case for `None` as a real value: #128

ramnes:
I think we should just write sensible defaults in `pick`, that would ignore `None` for `start_cursor` and maybe other variables. I don't think there's a need for control over both the presence and the value.

nicobako:
Yeah, this is a very tricky problem... one that I think will continue to haunt us forever!

I think your solution sounds good. The initial issue was that users wanted to type `start_cursor=None` (in order to make the Python code simpler).  My solution was to blindly ignore all *key-value* pair with a value of `None`.

Your approach sounds better. Just ignore specific *key-value* pairs with a value of `None` (like `start_cursor=None`), and keep all others.

So the final behavior of `pick` would be something like this:

```python
pick({start_cursor: None}, "start_cursor")

> {}

pick({"icon": None}, "icon")

> {"icon": None}
```

Right?

ramnes:
Yes, exactly. :)

nicobako:
Sounds good. One more annoying question...

Would reimplementing `pick` to something like this be good?

```python
def pick(base: Dict[Any, Any], *keys: str, *, keys_to_pop_if_none:List[Any]) -> Dict[Any, Any]:
    """Return a dict composed of key value pairs for keys passed as args."""
    # first, get all key-values of interest from base
    picked = {key: base[key] for key in keys if key in base}
    # next, pop any key-values if None is not an acceptable value
    for key in keys_to_pop_if_none:
        if picked[key] is None:
            picked.pop(key)
    
    return picked
```

This allows code that calls the `pick` function to be more specific about which keys to keep, and which to eliminate based on whether the value is `None`.

I am only suggesting this because I can envision a situation where a specific key can be `None` in some cases, but not `None` in others. One potential example is that, when *updating* a *block* a value of `None` might be okay, but when *querying* a *block* a value of `None` might not okay.

In the official notion-sdk, I can see a place where [`icon` can have a value of `null`](https://github.com/makenotion/notion-sdk-js/blob/main/src/api-endpoints.ts#L4711), and a place where [`icon` cannot have a value of `null`](https://github.com/makenotion/notion-sdk-js/blob/main/src/api-endpoints.ts#L8102). So, in some cases we may want to allow Python-users to type `icon=None`, and it be silently removed by `pick`, but other times when `icon=None` is kept.

I may be overthinking and worrying about this too much... Let me know what you think.

ramnes:
The problem we had so far was just with `start_cursor`, so I was thinking of something much simpler like:

```python
def pick(base: Dict[Any, Any], *keys: str) -> Dict[Any, Any]:
    """Return a dict composed of key value pairs for keys passed as args."""
    result = {}
    for key in keys:
        value = base.get(key)
        if value is None and key == "start_cursor":
            continue
        entries[key] = value
    return result
```

Eventually we could add the `keys_to_pop_if_none` variable if there's a need for other keys than `start_cursor`, and ultimately make it a function parameter if it's needed by someone, but again, let's wait to see these scenarios. (In any case, I'd give it a simpler name such as `drop_none`)

nicobako:
I think you're right. I'm always overcomplicating things...

I'll do my best get a pull request with the fixes, and some accompanying tests soon.

ramnes:
Alright so actually there are a lot of keys that can't be `null`, not just `start_cursor`, see the diff in 28020cd. :P

My first thought was to do the opposite and list the keys that *can* be `null`, but doing it the way I did will make maintenance easier since the API will yell at us if we try to send `null` when we're not allowed to.

ramnes:
Actually the code I wrote earlier is dumb because it adds all base values as `None`. There's a much better way: consider arguments only when they are handed over and simply ignore the infamous `start_cursor=None`... Implemented in 3289bd4.

PR #102, opened 2022-01-11
added test to ensure that key-value pairs with value of None get removed

nicobako:
@ramnes , I forgot to add a test to ensure that this new `pick` functionality is tested.

P.S. I am having a hard time setting up my _pre-commit_ to get all of the tests to pass. Was wondering if you could give me more details on how to do that. Thanks.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -4,12 +4,18 @@ from notion_client.helpers import get_id, get_url, pick
 
 
 def test_pick():
-    my_dict = {"Product": "Notion", "API": 2021, "python-sdk": "ramnes"}
+    my_dict = {
+        "Product": "Notion",
+        "API": 2021,
+        "python-sdk": "ramnes",
+        "optional-variable": None,
+    }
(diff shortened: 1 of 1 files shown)


ramnes:
Looking good, thanks!

What problem do you get with pre-commit exactly?

nicobako:
Actually, I got the _pre-commit_ to work... I'm using _wsl_ on Windows and just had to install gcc.

I don't know the `NOTION_TOKEN` environment variable used in the tests... not sure if that's something you want to share with others though.

ramnes:
You don't need one as long as you're using the cassettes (records of the API results that are saved in `tests/cassettes/`), which should be the default if you've installed `requirements/dev.txt`.

`NOTION_TOKEN` is only useful when you want to record a new cassette or regenerate previous ones. And any default Notion workspace should work IIRC!

ramnes:
I just added a commit that should help, is it better now? :)

nicobako:
> I just added a commit that should help, is it better now? :)

Awesome! Everything passes! Thanks!

PR #115, opened 2022-02-25
Added conftest type hints

sudosubin:
I'm not sure this test case would be helpful. Feel free to comment.

- Add small type hints for `conftest.py`
- ~~Add a logging module test case (`test_make_console_logger`)~~ -> Reverted
- Add pytest configuration (`asyncio_mode = auto`) to suppress `DeprecationWarning` when running pytest

  ```sh
  pytest_asyncio/plugin.py:191: DeprecationWarning: The 'asyncio_mode' default value will change to 'strict' in future, please explicitly use 'asyncio_mode=strict' or 'asyncio_mode=auto' in pytest configuration file.
      config.issue_config_time_warning(LEGACY_MODE, stacklevel=2)
  ```


tests/test_logging.py
```diff
@@ -0,0 +1,17 @@
...
+
+    with caplog.at_level(logging.INFO, logger="notion_client"):
+        logger.info("TEST_LOGGER_INFO_MESSAGE")
+        assert caplog.records[0].message == "TEST_LOGGER_INFO_MESSAGE"
+        caplog.clear()
+
+    with caplog.at_level(logging.DEBUG, logger="notion_client"):
+        logger.debug("TEST_LOGGER_DEBUG_MESSAGE")
+        assert caplog.records[0].message == "TEST_LOGGER_DEBUG_MESSAGE"
+        caplog.clear()
```
ramnes: Can we test our actual logging calls rather than the logging library itself? 😅

ramnes (changes requested):
I'm good with the first two commits, feel free to throw them into another PR if you want them merged quickly – or see my comment for the third commit. :)

tests/test_logging.py
sudosubin: I agree with your comment :)
It seems better to have `Client.request`, `Async.request` methods tested in this way.

sudosubin: I will revert the third commit from this PR, and will open new PR for logging test :)

ramnes:
Thanks!

PR #116, opened 2022-02-26
Added api endpoints test cases (blocks, blocks_children)

sudosubin:
I made a subdirectory `./tests/api_endpoints/` for `./notion_client/api_endpoints.py` test cases, since there are amount of classes.

Any comments are welcome.

- Added `BlocksEndpoint`, `BlocksChildrenEndpoint` test case
- Some request/response come from [notion api document](https://developers.notion.com/reference/intro).

### DIFF
--- a/tests/api_endpoints/cassettes/test_endpoint_blocks_children_append.yaml
+++ b/tests/api_endpoints/cassettes/test_endpoint_blocks_children_append.yaml
@@ -1,0 +1,42 @@
+interactions:
+- request:
+    body: '{"children": [{"object": "block", "type": "heading_2", "heading_2": {"text":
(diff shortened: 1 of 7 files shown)


ramnes:
Sorry for the long answer but I'm not sure what I want to do with the tests. On this PR specifically, I don't like the fact that the block ids are set in stone, and that since you are the one that generated the cassettes, I wouldn't be able to regenerate them myself with my own API key and Notion workspace. Do you have any idea on what we could do so that it's a bit more permissive?

ramnes:
Closing because there is no answer from OP but feel free to answer with any hint regarding my previous comment and I will gladly reopen, I'm definitely interested in more tests. :)

PR #118, opened 2022-03-08
Use PEP 517 build system

CyberTailor:
This includes pyproject.toml file and declarative setup.cfg config.


ramnes:
Thanks for the PR @CyberTailor!

Can you comment in #5 on what this brings to the library users and to myself (as a maintainer)?

setup.cfg
```diff
@@ -1,36 +1,46 @@
...
+	Programming Language :: Python :: 3.9
+	Programming Language :: Python :: 3.10
+	Development Status :: 4 - Beta
+	License :: OSI Approved :: MIT License
+
+[options]
+packages = notion_client
+
+[options.package_data]
+    notion_client = py.typed
```
ramnes: Are the leading whitespace intentional?

setup.cfg
```diff
@@ -1,36 +1,46 @@
...
+install_requires = httpx >= 0.15.0
+classifiers=
+	Programming Language :: Python
+	Programming Language :: Python :: 3
+	Programming Language :: Python :: 3.7
+	Programming Language :: Python :: 3.8
+	Programming Language :: Python :: 3.9
+	Programming Language :: Python :: 3.10
+	Development Status :: 4 - Beta
+	License :: OSI Approved :: MIT License
```
ramnes: Please stay consistent by using spaces instead of tabs, and ideally please add a TOML linter in `.pre-commit-config.yaml`. :)

ramnes (commented):
LGTM overall on the "code" side, just a few nitpicks. I still need to be convinced in #5 tho. :)

ramnes:
Sorry but I'm going to put this aside for now. setup.cfg isn't required anymore since a month or so, so this PR is already outdated. My point is that the Python packaging world is still evolving a lot and I don't want to spend time on packaging until it's settled on something more stable. I know you don't want to hear this but I feel that setup.py is still is the defacto standard, and I don't want to be the one smoothing out the rough edges. I'll consider this again in a year or two. :)

Issue #139, opened 2022-07-21
Support for comments API


ramnes:
Let's port this: https://github.com/makenotion/notion-sdk-js/pull/326

ramnes:
Implemented in 6aff9f37b24182c9701b7e7c5b0bd300cb0208e3

PR #147, opened 2022-09-11
added comments endpoint and updated mypy

mutt0-ds:
Hello! 

I've added the comments endpoint (following Notion's structure) and closing #139 and also updated mypy in the pre-commit file, there was a bug with my newer Python version (see [here](https://github.com/python/mypy/issues/5787)).
Thanks as always for the library :)


ramnes:
Cool, thanks!

I just merged it manually to restore the YAML indentation and split the commit. :)

Issue #137, opened 2022-07-19
Implement notion-sdk-js helpers


ramnes:
Notion now implements a few helpers: https://github.com/makenotion/notion-sdk-js/pull/320

Let's port them back here with the same API!

ramnes:
@jheddings You may be interested in these, as a higher-level base for notional. :)

ramnes:
The README examples are also to port back along the way: https://github.com/makenotion/notion-sdk-js/pull/325

mutt0-ds:
It looks interesting, I'll be happy to contribute!

PR #148, opened 2022-09-14
Added helper functions (see #137)

mutt0-ds:
Hello! I've worked on the issue #137 and added the helper functions following the same structure of notion-sdk-js.

I have a couple of doubts about the typing in the Async version, since I'm rarely working with this kind of functions, so please let me know if it's ok.

I tested them locally with my Notion. but I couldn't add tests because they require a mock call to the API... Should we raise an issue and work on it?

We also need to update the README, I'll do it if the request is approved

### DIFF
--- a/notion_client/helpers.py
+++ b/notion_client/helpers.py
@@ -37,7 +37,7 @@ def iterate_paginated_api(
         yield response.get("results")
 
         next_cursor = response.get("next_cursor")
-        if (not response["has_more"]) | (next_cursor is None):
+        if not response.get("has_more") or not next_cursor:
             return
 
 
--- a/notion_client/helpers.py
+++ b/notion_client/helpers.py
@@ -27,13 +27,13 @@ def get_id(url: str) -> str:
 
 
 def iterate_paginated_API(
-    function: Callable[..., Any], first_page_args: Dict[Any, Any]
+    function: Callable[..., Any], **kwargs: Any
 ) -> Generator[List[Any], None, None]:
     """Return an iterator over the results of any paginated Notion API."""
     next_cursor = None
 
     while True:
-        response = function(**first_page_args, start_cursor=next_cursor)
+        response = function(kwargs, start_cursor=next_cursor)
         yield response.get("results")
 
         next_cursor = response.get("next_cursor")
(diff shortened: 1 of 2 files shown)


notion_client/helpers.py
```diff
@@ -24,3 +24,78 @@ def get_id(url: str) -> str:
         raise ValueError("The path in the URL seems to be incorrect.")
     raw_id = path[-32:]
     return str(UUID(raw_id))
+
+
+def iterate_paginated_API(
+    function: Callable[..., Any], first_page_args: Dict[Any, Any]
```
ramnes: Let's use kwargs?

```suggestion
    function: Callable[..., Any], **kwargs: Any
```

notion_client/helpers.py
```diff
@@ -24,3 +24,78 @@ def get_id(url: str) -> str:
         raise ValueError("The path in the URL seems to be incorrect.")
     raw_id = path[-32:]
     return str(UUID(raw_id))
+
+
+def iterate_paginated_API(
```
ramnes: IIRC Python functions must always be lowercase

```suggestion
def iterate_paginated_api(
```

notion_client/helpers.py
```diff
@@ -24,3 +24,78 @@ def get_id(url: str) -> str:
...
+) -> Generator[List[Any], None, None]:
+    """Return an iterator over the results of any paginated Notion API."""
+    next_cursor = None
+
+    while True:
+        response = function(**first_page_args, start_cursor=next_cursor)
+        yield response.get("results")
+
+        next_cursor = response.get("next_cursor")
+        if (not response["has_more"]) | (next_cursor is None):
```
ramnes:
```suggestion
        if not response.get("has_more") or not next_cursor:
```

ramnes (changes requested):
LGTM overall, just a few little things to change

ramnes:
Thank you very much for taking this work!

I would love if the README changes could be added to the same commit as well. :)

For the tests, it's true that cassettes are great for me as a sole-developer but not practical at all for collaboration. Do you have any idea on how we could do things differently? c.f. https://github.com/ramnes/notion-sdk-py/pull/116#issuecomment-1101463557

mutt0-ds:
I'll work on the README tomorrow, while I'm thinking about an easy way for handling tests. We can mockup all the responses or maybe use a secret token associated to the notion example at [test_helpers](https://github.com/ramnes/notion-sdk-py/blob/main/tests/test_helpers.py)?

ramnes:
I think the key is to not rely on existing pages or other data we can't control. We should just give the tests a private token and a root page, and then they'd bootstrap a test environment (i.e. a few pages) within the given root page.

That way, anyone could run the tests with their own Notion account. We could still use the cassettes to speed up most tests, but that wouldn't be a blocker anymore to add more tests.

notion_client/helpers.py
ramnes: Did you resolve it but forgot to include the changes in the commit? :sweat_smile:

notion_client/helpers.py
```diff
@@ -24,3 +24,76 @@ def get_id(url: str) -> str:
...
+
+
+def iterate_paginated_API(
+    function: Callable[..., Any], **kwargs: Any
+) -> Generator[List[Any], None, None]:
+    """Return an iterator over the results of any paginated Notion API."""
+    next_cursor = None
+
+    while True:
+        response = function(kwargs, start_cursor=next_cursor)
```
evertheylen: Is this not supposed to be `**kwargs`? Awaiting the merge of this PR, I copied the code but I had to change this to make it work.

mutt0-ds:
@ramnes OK about tests, I agree. 
I'll think about the setup in the next days (and probably open a new issue or referring to the existing one about coverage). Thanks!

davideph:
> LGTM overall, just a few little things to change

Changes done!

notion_client/helpers.py
ramnes: Did you forget to commit these changes?

ramnes:
One little thing left :)

ramnes:
Thanks! 🚀

PR #152, opened 2022-09-28
Introduce tests for API endpoints

mutt0-ds:
Hi!
I've worked on the tests for the library's endpoints, I think that this could be a good starting point for extendending the coverage.
In short, I created a Notion testing account with a connection, retrieved the NOTION_TOKEN, created a testing page and set its url as NOTION_TESTING_URL.

The tests are using NOTION_TESTING_URL for creating a temporary subpage with the current timestamp as name, and it will be used for a single run (and then deleted afterwards). Inside, the first tests are creating new objects (a DB, a comment, a block and a page), then the others are goint to test all the available functions. Ignoring the deprecated function databases.list, I achieved a 97% coverage of the module.

@ramnes I can send you the secrets I used in my example, but you can run the tests by yourself with your NOTION_TOKEN and a page in your Notion account (the bot must have full access to all the capabilities). You will notice that a tempporary page will be created inside NOTION_TESTING_URL, and it will disappear at the end of the testing run.

I'm open to feedbacks, this was quite an adventure to me.


mutt0-ds:
If this works:
- Add the secrets to the Github actions (rn the automatic tests are failing because there is no URL and TOKEN!)
- Update README with instructions for local testing
- Add tests for helpers

ramnes:
Just took a quick look and that looks nice! I'm about to leave a few comments but that's not a full review as I'm quite busy right now, please bear with me if I take a few days to review completely the PR.

tests/conftest.py
```diff
@@ -19,11 +19,24 @@ def remove_headers(response: dict):
...
+def token() -> str:
+    notion_token = os.environ.get("NOTION_TOKEN")
+    if not notion_token:
+        raise EnvironmentError("Env variable NOTION_TOKEN needed for testing")
+    return notion_token
 
 
-@pytest.fixture
+@pytest.fixture(scope="session")
+def testing_page() -> str:
```
ramnes: Let's make the fixture name and environment variable name consistent so that it's easier to read and understand. :)

tests/test_endpoints.py
```diff
@@ -0,0 +1,236 @@
...
+
+# PAGES
+# `test_pages_create` generates a temporary page and uses its ID
+# as input for all the other tests. Their objects (databases, blocks, comments)
+# will be created inside the page. With this method we have a
+# unique environment for each run, using the current datetime as identifier
+
+
+@pytest.fixture(scope="session")
+def test_pages_create(client, testing_page) -> str:
```
ramnes: I'd prefer that we keep the `test_` prefix for test functions, and use something that looks like a variable name for fixtures.

tests/test_endpoints.py
```diff
@@ -0,0 +1,236 @@
...
+
+def compare_cleaned_ids(id1: str, id2: str) -> bool:
+    return id1.replace("-", "").strip() == id2.replace("-", "").strip()
+
+
+# PAGES
+# `test_pages_create` generates a temporary page and uses its ID
+# as input for all the other tests. Their objects (databases, blocks, comments)
+# will be created inside the page. With this method we have a
+# unique environment for each run, using the current datetime as identifier
```
ramnes: I'm not too fond of such comments, I can read the code instead :P

tests/conftest.py
ramnes: Also, is it really a URL or just the page id?

mutt0-ds: A page id, I'll rename it as NOTION_TESTING_PAGE_ID for clarity

mutt0-ds:
Thanks for all the feedbacks and no problem, I will need some time too for correcting! 
I'll update names and comments as suggested and create new fixtures that call the tests and yield the ID of the result

tests/conftest.py
ramnes: LGTM (although I think you can drop the ING :P)

mutt0-ds:
@ramnes any update on the PR? I committed the changes you have requested :)

tests/test_endpoints.py
```diff
@@ -0,0 +1,223 @@
+from datetime import datetime
+
+import pytest
+
+TEST_PAGE_NAME = f"Test Page - {datetime.now()}"
+
+
+# SETUP
```
ramnes: All things related to the setup should go in the conftest I think :)

tests/test_endpoints.py
```diff
@@ -0,0 +1,223 @@
...
+def session_block_id(client, session_page_id) -> str:
+    children = [
+        {"paragraph": {"rich_text": [{"text": {"content": "I'm a paragraph."}}]}}
+    ]
+
+    response = client.blocks.children.append(
+        block_id=session_page_id, children=children
+    )
+    assert response["type"] == "block"
+    assert response["results"][0]["type"] == "paragraph"
```
ramnes: Why do we need asserts in a fixture?

tests/test_endpoints.py
```diff
@@ -0,0 +1,223 @@
...
+            "title": [{"text": {"content": TEST_PAGE_NAME}}],
+        },
+        # using children = [] for avoiding issue with the Notion API
+        # when creating a subpage without blocks
+        children=[],
+    )
+
+    yield response["id"]
+
+    # at the end of the session, delete the page
```
ramnes: The comment doesn't feel necessary to me :)

tests/test_endpoints.py
```diff
@@ -0,0 +1,223 @@
...
+    assert response["rich_text"][0]["plain_text"] == rich_text[0]["text"]["content"]
+
+    return response["id"]
+
+
+def compare_cleaned_ids(id1: str, id2: str) -> bool:
+    return id1.replace("-", "").strip() == id2.replace("-", "").strip()
+
+
+# PAGES
```
ramnes: It seems to me that the way you named the tests is quite clear and consistent, so these comments don't feel necessary to me! I understand the appeal but I prefer to make my code files as small as possible. :)

tests/test_endpoints.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+def test_pages_retrieve(client, session_page_id, page_id):
```
ramnes: Let's keep using pytest-vcr and commit the cassettes for these tests so that we (or the CI) can run the them faster and without a NOTION_TOKEN.

ramnes (approved):
A few bits I would prefer differently but LGTM overall, thank you very much @mutt0-ds and sorry for the late review! I can do the work myself, tell me if you want to address the comments or if you'd prefer me to do so, for any reason. :)

mutt0-ds:
@ramnes I managed to use cassettes as requested, and cleaned up the comments a bit!

Notable updates:
- Fixture don't work with the pytest.mark.vcr() decorator so I separated them: the cassette runs the test, and then the fixture opens up the YAML file and read the ID to be used for the other tests
- I had to set up a custom matcher as some HTTP requests contains the ID of the page/block/database used and can interfere if someone else is testing a new cassette with a different environment
- A NOTION_ID_TOKEN is not required anymore for testing unless you want to create new cassettes

Let me know if everything looks good to you :)

cassettes/test_pages_create.yaml
```diff
@@ -0,0 +1,113 @@
...
+- request:
+    body: '{"parent": {"page_id": null}, "properties": {"title": [{"text": {"content":
+      "Test Page - 2022-11-13 11:58:44.392249"}}]}, "children": []}'
+    headers:
+      accept:
+      - '*/*'
+      accept-encoding:
+      - gzip, deflate
+      authorization:
+      - Bearer secret_UWDjvhuUx693qHExqJxQ9nu1KabHlyBJ78u3vUM8YDF
```
ramnes: You probably want to renew your secret now that it has leaked, and then change the code so that it doesn't appear anymore :)

tests/test_endpoints.py
```diff
@@ -0,0 +1,232 @@
...
+import pytest
+
+from tests.conftest import TEST_PAGE_NAME
+
+
+@pytest.fixture(scope="module")
+def session_page_id(vcr) -> str:
+    with vcr.use_cassette("test_pages_create.yaml") as cass:
+        response = cass._serializer.deserialize(cass.data[0][1]["content"])
+        return response["id"]
```
ramnes: What is the difference between this fixture and `conftest.page_id`?

tests/test_endpoints.py
```diff
@@ -0,0 +1,232 @@
...
+        "rich_text": [
+            {
+                "text": {"content": new_plain_text},
+                "annotations": {"bold": True, "color": "red_background"},
+            }
+        ]
+    }
+    response = client.blocks.update(block_id=session_block_id, paragraph=new_text)
+
+    assert response["paragraph"]["rich_text"][0]["plain_text"] == new_plain_text
```
ramnes: If we've already set that value in a previous test run, isn't the current test run going to pass even if the update failed?

cassettes/test_pages_create.yaml
mutt0-ds: wooooops all those cassettes confused me 😖 luckily it was just an empty environment specifically for those tests. My bad!

tests/test_endpoints.py
mutt0-ds: I added a couple of comments and moved the fixtures in conftest for clarity. 
In short, for creating the session_page_id, which is the page where the tests will run, we need a parent (see test_pages_create), so it's necessary to use both of them. The only function of page_id is to be a parent in the API call.
Feedbacks are welcomed for a clearer name :)

tests/test_endpoints.py
mutt0-ds: Yes, this is not going to work if we are using cassettes... I had to clean up the assertions because we can't compare the value
Again, feedbacks are welcomed about the best way to test these endpoints

tests/conftest.py
```diff
@@ -15,15 +20,70 @@ def remove_headers(response: dict):
...
+
+
 @pytest.fixture(scope="session")
-def token() -> Optional[str]:
-    return os.getenv("NOTION_TOKEN")
+def token() -> str:
+    notion_token = os.environ.get("NOTION_TOKEN")
+    if not notion_token:
+        raise EnvironmentError("Env variable NOTION_TOKEN needed for testing")
+    return notion_token
```
ramnes: It shouldn't be needed if we use the cassettes :P

tests/test_endpoints.py
ramnes: well, we probably want to make the tests autonomous, e.g. with a function-scoped fixture that creates a page just for that one specific test – AFAICT that's almost what you're doing already, but a function-scoped fixture would guarantee that no other test leak into this one

tests/test_endpoints.py
ramnes: Makes sense! Maybe we could rename the environment variable and fixture from  `NOTION_TEST_PAGE_ID` to `PARENT_PAGE_ID`? We could even drop the `session_` prefix if we go that way, `parent_page_id` and `page_id` seem clear enough to me.

tests/test_endpoints.py
mutt0-ds: Great idea! I transformed page_id into a function-scoped fixture that generates a new page for each single test- If necessary (with blocks tests and databases tests), it also temporarily generates a block_id or database_id inside the page!
Now it works :)

ramnes (approved):

ramnes:
Thank you very much for your work on this @mutt0-ds and so sorry for the wait!

PR #153, opened 2022-10-10
Explicitly use `notion_client` as package to avoid exporting `tests`

tjhoff:
Using `find_packages` in setup.py also exports `tests`, as that has a `__init__.py`. 

```
>>> from setuptools import find_packages
>>> find_packages()
['tests', 'notion_client']
```

This causes issues when `notion-client` is used in projects with pylint that contain a `tests` directory, as pylint prefers the installed `tests` package from notion-client over the local discovered `tests` directory.

An alternative approach would be to remove the `__init__.py`, but I'm not sure if it's there for a specific reason. 

LMK if you need any more info - I've run through the pre-commit hooks and it all looks good.

Thanks for the great project - does the job well.

### DIFF
--- a/notion_client/client.py
+++ b/notion_client/client.py
@@ -89,7 +89,7 @@ class BaseClient:
         client.headers = httpx.Headers(
             {
                 "Notion-Version": self.options.notion_version,
-                "User-Agent": "ramnes/notion-sdk-py@1.0.0",
+                "User-Agent": "ramnes/notion-sdk-py@1.0.1",
             }
         )
         if self.options.auth:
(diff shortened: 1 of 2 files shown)


ramnes:
Cool, thanks!

ramnes:
Since there are a few unreleased commits behind yours, I won't make it a tag or release here and I'm thus amending the commit to put back 1.0.0, but I still pushed that 1.0.1 on PyPI if you want to use it. :)

tjhoff:
Oops, sorry about that, didn't do my due diligence there. Thanks much!

PR #168, opened 2022-12-10
100% Coverage 🥳

mutt0-ds:
Hello! I added some tests for the errors and fixed a couple of minor issues (TESTING_PAGE_NAME became a session fixture, I excluded abstract properties from tests and I used "with" statements to test the __enter__ and __exit__ properties of the Client class). 

When testing the errors I used [https://httpstat.us/](https://httpstat.us/) for simulating HTTP errors, let me know if this solution would be OK as I didn't manage to generate this kind of situation with cassettes only. :)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,4 @@
+[report]
+exclude_lines =
+    pragma: no cover
+    @abstract
--- a/notion_client/client.py
+++ b/notion_client/client.py
@@ -120,7 +120,7 @@ class BaseClient:
             try:
                 body = error.response.json()
                 code = body.get("code")
-            except json.JSONDecodeError:
+            except json.JSONDecodeError:  # pragma: no cover
                 code = None
             if code and is_api_error_code(code):
                 raise APIResponseError(response, body["message"], code)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,8 +8,6 @@ import pytest
 
 from notion_client import AsyncClient, Client
 
-TEST_PAGE_NAME = f"Test Page - {datetime.now()}"
-
 
 @pytest.fixture(scope="session")
 def vcr_config():
@@ -37,6 +35,11 @@ def vcr(vcr):
 @pytest.fixture(scope="session")
 def token() -> str:
     return os.environ.get("NOTION_TOKEN")
+
+
+@pytest.fixture(scope="session")
+def test_page_name():
+    return f"Test Page - {datetime.now()}"
 
 
 @pytest.fixture(scope="module")
@@ -119,13 +122,15 @@ def comment_id(client, page_id) -> str:
     yield response["id"]
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="module")
 def client(token: Optional[str]):
-    return Client({"auth": token})
+    with Client({"auth": token}) as client:
+        yield client
+    client.close()
 
 
 @pytest.fixture
 async def async_client(token: Optional[str]):
-    client = AsyncClient({"auth": token})
-    yield client
+    async with AsyncClient({"auth": token}) as client:
+        yield client
     await client.aclose()
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -49,5 +49,3 @@ async def test_async_client_request_auth(token):
 
     response = await async_client.request("/users", "GET", auth=token)
     assert response["results"]
-
-    await async_client.aclose()
--- a/tests/test_endpoints.py
+++ b/tests/test_endpoints.py
@@ -1,16 +1,12 @@
-from datetime import datetime
-
 import pytest
-
-from tests.conftest import TEST_PAGE_NAME
 
 
 @pytest.mark.vcr()
-def test_pages_create(client, parent_page_id):
+def test_pages_create(client, parent_page_id, test_page_name):
     response = client.pages.create(
         parent={"page_id": parent_page_id},
         properties={
-            "title": [{"text": {"content": TEST_PAGE_NAME}}],
+            "title": [{"text": {"content": test_page_name}}],
         },
         children=[],
     )
@@ -124,8 +120,8 @@ def test_search(client, page_id):
 
 
 @pytest.mark.vcr()
-def test_databases_create(client, page_id):
-    db_name = f"Test Database - {datetime.now()}"
+def test_databases_create(client, page_id, test_page_name):
+    db_name = f"Test Database - {test_page_name}"
     properties = {
         "Name": {"title": {}},  # required property
     }
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -1,0 +1,57 @@
+import pytest
+
+from notion_client import AsyncClient, Client
+from notion_client.errors import (
+    APIResponseError,
+    HTTPResponseError,
+    RequestTimeoutError,
+    is_api_error_code,
+)
+
+STATUS_PAGE_BAD_REQUEST = "https://httpstat.us/400"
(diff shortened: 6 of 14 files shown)


tests/test_client.py
```diff
@@ -42,12 +42,10 @@ def test_client_request_auth(token):
...
-    async_client = AsyncClient()
+    async_client = AsyncClient(auth="Invalid")
 
     with pytest.raises(APIResponseError):
         await async_client.request("/users", "GET")
 
     response = await async_client.request("/users", "GET", auth=token)
     assert response["results"]
-
-    await async_client.aclose()
```
ramnes: Why removing that line?

tests/conftest.py
```diff
@@ -119,13 +122,15 @@ def comment_id(client, page_id) -> str:
...
+    client.close()
 
 
 @pytest.fixture
 async def async_client(token: Optional[str]):
-    client = AsyncClient({"auth": token})
-    yield client
+    async with AsyncClient({"auth": token}) as client:
+        yield client
     await client.aclose()
```
ramnes: Why do we still need to manually close the clients if we now use the context manager? What's the point of using the context manager?

.coveragerc
```diff
@@ -0,0 +1,4 @@
+[report]
+exclude_lines =
+    pragma: no cover
+    @abstract
```
ramnes, line 4: Why is it needed?

notion_client/client.py
```diff
@@ -120,7 +120,7 @@ def _parse_response(self, response: Response) -> Any:
             try:
                 body = error.response.json()
                 code = body.get("code")
-            except json.JSONDecodeError:
+            except json.JSONDecodeError:  # pragma: no cover
```
ramnes: Why not testing this line?

tests/test_endpoints.py
```diff
@@ -124,8 +120,8 @@ def test_search(client, page_id):
 
 
 @pytest.mark.vcr()
-def test_databases_create(client, page_id):
-    db_name = f"Test Database - {datetime.now()}"
+def test_databases_create(client, page_id, test_page_name):
+    db_name = f"Test Database - {test_page_name}"
```
ramnes: Why is it better that way?

tests/test_errors.py
```diff
@@ -0,0 +1,61 @@
...
+
+@pytest.mark.vcr()
+def test_api_response_error(client):
+    with pytest.raises(APIResponseError):
+        client.request("/invalid", "GET")
+    with pytest.raises(APIResponseError):
+        client.request("/users", "GET", auth="Invalid")
+
+
+def test_api_request_timeout_error(token):
```
ramnes: No cassette here?

ramnes (commented):
Curious about a few things but that's awesome we're getting there!

tests/test_endpoints.py
mutt0-ds: Just for convenience I removed the datetime module import and just used the fixture which already contains the current datetime. Totally an aesthetic choice, I didn't like to use the import just for that single line

tests/test_errors.py
mutt0-ds: So, that's a tricky one. The test makes a request but there is no response as I artificially create a timeout, so the cassette is not created. I don't like it because if there is no connection the test will fail, let me know if there is any workaround

.coveragerc
mutt0-ds: To exclude abstract classes from coverage (found [here](https://stackoverflow.com/a/41284681)) since they can't be tested. In this case the "request" function from BaseClient.

tests/conftest.py
mutt0-ds: I used the context manager to test both the "__enter__" and "__exit__" and the "close" method. Honestly I never used the client with a context manager so this is purely for coverage reasons. We can split the tests in two: one is using a classic approach and another one is using the context manager, for clarity?

tests/test_client.py
mutt0-ds: I removed aclose because I didn't want to have a closed client to interfere with the other tests, and the line was already covered in conftest (async_client). Probably it is unnecessary but I like to keep both sync and async tests the same, and if I close the sync client the other tests were failing (being a module fixture). 
Let me know if the explanation makes sense aha :)

notion_client/client.py
mutt0-ds: Because I can't generate a request from the Notion API with uncorrect JSON data to test. Help is welcomed with this specific case

.coveragerc
ramnes: Makes sense!

notion_client/client.py
ramnes: Can't we just test it with http://httpstat.us/200? It doesn't return JSON.

tests/conftest.py
ramnes: Ah! Well, ideally we'd want the context manager methods and close/aclose methods to have their own tests, but I guess we can just remove the close/aclose calls here and call it done for now. :)

tests/test_client.py
ramnes: Mhh, in this test and the one just above, client and async_client aren't fixture, they are instantiated within the tests. That's why we need to close them, otherwise they're not going to be closed ever and that could give us some warning at the ends of the tests.

tests/test_endpoints.py
ramnes: I'm not fond of the final name ("Test Database - Test Page - date"). We can simply remove the datetime logic entirely and call the database like the test, i.e. `test_databases_create`. The page that we use as a parent (with `page_id`) already has a unique name based on the datetime so no clash would be possible if we use the test name. What do you think?

tests/test_endpoints.py
```diff
@@ -1,16 +1,12 @@
...
-
 
 @pytest.mark.vcr()
-def test_pages_create(client, parent_page_id):
+def test_pages_create(client, parent_page_id, test_page_name):
     response = client.pages.create(
         parent={"page_id": parent_page_id},
         properties={
-            "title": [{"text": {"content": TEST_PAGE_NAME}}],
+            "title": [{"text": {"content": test_page_name}}],
```
ramnes: Same thing here than [my other comment below](https://github.com/ramnes/notion-sdk-py/pull/168/files#r1045298165), by the way! We could just call the page `test_pages_create` if we go that way. Am I missing something?

tests/test_errors.py
ramnes: If there is no connection then it's going to timeout as well I guess! That probably good enough.

tests/test_endpoints.py
mutt0-ds: I agree with you, the parent page should be enough. Just calling them Test Database then :)

tests/test_client.py
mutt0-ds: All clear now! Fixed it

notion_client/client.py
mutt0-ds: Nevermind, it's being covered, I removed the # no cover comment :)

mutt0-ds:
So, I applied the minor tweaks but I am not 100% happy with the timeout test, as you can see a Github action failed for a connection issue (I guess)...

ramnes:
Yep, I just ran the actions again and it does look flaky, not sure why. I'd say that's probably a good place for a mock?

mutt0-ds:
I simulated a timeout with monkeypatch raising a TimeoutException, this should do the trick. Codecov doesn't look too happy but from the errors it looks like a token issue

ramnes:
Thank you once again @mutt0-ds, it's merged! 🎉

mutt0-ds:
It's been a pleasure @ramnes, I've learned so much by working at your project! Let me know if I can contribute to other things :)

PR #181, opened 2023-03-18
Add filter_properties in databases query

sgskt:
Adds the possibility to specify properties_id in the databases query endpoint

image


ramnes:
Thanks!

PR #195, opened 2023-05-07
multipage support for databases.query

moarsoarse:
Sorry for clunky syntax, just started learning Python. But it works))


ramnes:
Hey there,

I won't merge this because there's already the `iterate_paginated_api` helper, as mentioned in the README and the docs. This is what you should use. :)

Others reasons are that I don't want to diverge from the JavaScript SDK API and that I don't think your diff will work with the asynchronous client.

Thanks for the effort though!

PR #198, opened 2023-06-29
Implement is_full_page_or_database

ramnes:
https://github.com/makenotion/notion-sdk-js/pull/413 port


ramnes:
@mutt0-ds I don't really like that I have to rebuild all cassettes since I don't have the same API key and parent page id as yours. Do you have any idea on how we could avoid this?

mutt0-ds:
@ramnes my suggestion is that you regenerate them with a key and parent page id on your own and eventually handle the official credentials to contributors (myself included), like you have done in the PR. My test was made with a disposable account because I wanted to use cassettes locally, independently from who was running the tests. 

Sure, committing all of them each time is annoying but I don't see how to bypass the issue, maybe by masking the data that changes in the yaml such as ID and time, but it will require more time and complexity than the benefit.

Issue #199, opened 2023-08-08
Feature Request: add types for all responses using `TypedDict`


guacs:
I was wondering if there is any interest in specifying the types of each response instead of just using `Any`. The types could be taken from the official API written in Typescript. I looked through #9 and so instead of using dataclasses, I'm proposing type hinting everything using `TypedDict`. This would make things quite easy for the users since they would get features like autocomplete and would be able run their code through a type checker. Essentially I'm suggesting something like the following:

```python
class Database(TypedDict):
    ...

class DatabasesEndpoint(Endpoint):

    def retrieve(self, db_id: str) -> Database:
        ...
```

However, one possible issue I'm seeing with this approach is that the return types of the endpoints currently are not suitable for this as far as I know. The `SyncAsync[T]` does not work well with type checkers so instead separate endpoints classes for synchronous and asynchronous endpoints would need to be maintained.

ramnes:
Yes, I'm definitely interested if you're willing to add them and most importantly maintain them afterwards to stay up to date with the TS client. :)

guacs:
Yupp I'm willing to add them and maintain them as well. I'll start working on it and hopefully if this makenotion/notion-sdk-js#294 gets resolved, then maintenance would be much easier.

superleesa:
@ramnes 

I think we can do something like this? I checked that this works for mypy without errors. This way, we can still have one class for each set of endpoints (no separate classes for sync and async clients). This uses two overloads for each endpoint instead.
```python
from __future__ import annotations
from typing import Generic, TypeVar, Any, TypedDict, Awaitable, Union, overload
from abc import ABC, abstractmethod


T = TypeVar('T', bound=Union[dict, Awaitable[dict]])
ResultDictType = TypeVar('ResultDictType')
MyDict = TypedDict('MyDict', {'name': str, 'age': int})


class Endpoint(Generic[T], ABC):
    def __init__(self, parent: BaseClient[T]) -> None:
        self.parent = parent


class MyDictOutputEndpoint(Endpoint[T]):
    @overload
    def get_something(self: MyDictOutputEndpoint[dict]) -> MyDict:
        ...
    
    @overload
    def get_something(self: MyDictOutputEndpoint[Awaitable[dict]]) -> Awaitable[MyDict]:
        ...
    
    def get_something(self) -> Awaitable[MyDict] | MyDict:
        return self.parent.request(method="get", cast_to=MyDict)


class BaseClient(Generic[T], ABC):
[...]

PR #200, opened 2023-08-24
Type Hint Responses [WIP]

guacs:
This is WIP PR for #199.

@ramnes, I just wanted to confirm whether this kind of an approach is fine or not. Also, I was thinking that instead of having to maintain a separate async and synchronous versions of the endpoints, we could use something like [unasyncd](https://github.com/provinzkraut/unasyncd).

Once this is confirmed, I'll go ahead and get to typing all the responses by comparing with the official docs and the official Javascript client.

### DIFF
--- a/notion_client/_async_api_endpoints.py
+++ b/notion_client/_async_api_endpoints.py
@@ -1,0 +1,308 @@
+# mypy: disable-error-code="no-any-return"
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, cast
+
+from notion_client.helpers import pick
+from notion_client.types import BotUserObjectResponse, UserObjectResponse
+
+if TYPE_CHECKING:
+    from typing import Any
+
+    from notion_client.client import AsyncClient
+
+
+class AsyncEndpoint:
+    def __init__(self, parent: AsyncClient) -> None:
+        self.parent = parent
+
+
+class BlocksChildrenEndpoint(AsyncEndpoint):
+    async def append(self, block_id: str, **kwargs: Any) -> Any:
+        """Create and append new children blocks to the block using the ID specified.
+
+        *[🔗 Endpoint documentation](https://developers.notion.com/reference/patch-block-children)*
+        """  # noqa: E501
+        return await self.parent.request(
+            path=f"blocks/{block_id}/children",
+            method="PATCH",
+            body=pick(kwargs, "children"),
+            auth=kwargs.get("auth"),
+        )
+
+    async def list(self, block_id: str, **kwargs: Any) -> Any:
+        """Return a paginated array of child [block objects](https://developers.notion.com/reference/block) contained in the block.
+
+        *[🔗 Endpoint documentation](https://developers.notion.com/reference/get-block-children)*
+        """  # noqa: E501
+        return await self.parent.request(
+            path=f"blocks/{block_id}/children",
+            method="GET",
+            query=pick(kwargs, "start_cursor", "page_size"),
+            auth=kwargs.get("auth"),
+        )
+
+
+class BlocksEndpoint(AsyncEndpoint):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self.children = BlocksChildrenEndpoint(*args, **kwargs)
+
+    async def retrieve(self, block_id: str, **kwargs: Any) -> Any:
+        """Retrieve a [Block object](https://developers.notion.com/reference/block) using the ID specified.
+
+        *[🔗 Endpoint documentation](https://developers.notion.com/reference/retrieve-a-block)*
+        """  # noqa: E501
+        return await self.parent.request(
+            path=f"blocks/{block_id}", method="GET", auth=kwargs.get("auth")
+        )
+
+    async def update(self, block_id: str, **kwargs: Any) -> Any:
+        """Update the content for the specified `block_id` based on the block type.
+
+        *[🔗 Endpoint documentation](https://developers.notion.com/reference/update-a-block)*
+        """  # noqa: E501
+        return await self.parent.request(
+            path=f"blocks/{block_id}",
+            method="PATCH",
+            body=pick(
+                kwargs,
+                "embed",
+                "type",
+                "archived",
+                "bookmark",
+                "image",
+                "video",
+                "pdf",
+                "file",
+                "audio",
+                "code",
+                "equation",
+                "divider",
+                "breadcrumb",
+                "table_of_contents",
+                "link_to_page",
+                "table_row",
+                "heading_1",
+                "heading_2",
+                "heading_3",
+                "paragraph",
+                "bulleted_list_item",
+                "numbered_list_item",
+                "quote",
+                "to_do",
+                "toggle",
+                "template",
+                "callout",
+                "synced_block",
+                "table",
+            ),
+            auth=kwargs.get("auth"),
+        )
+
+    def delete(self, block_id: str, **kwargs: Any) -> Any:
+        """Set a [Block object](https://developers.notion.com/reference/block), including page blocks, to `archived: true`.
+
+        *[🔗 Endpoint documentation](https://developers.notion.com/reference/delete-a-block)*
+        """  # noqa: E501
+        return self.parent.request(
+            path=f"blocks/{block_id}",
+            method="DELETE",
+            auth=kwargs.get("auth"),
+        )
+
+
+class DatabasesEndpoint(AsyncEndpoint):
+    async def list(self, **kwargs: Any) -> Any:  # pragma: no cover
+        """List all [Databases](https://developers.notion.com/reference/database) shared with the authenticated integration.
+
+        > ⚠️  **Deprecated endpoint**
+
+        *[🔗 Endpoint documentation](https://developers.notion.com/reference/get-databases)*
+        """  # noqa: E501
+        return await self.parent.request(
+            path="databases",
+            method="GET",
+            query=pick(kwargs, "start_cursor", "page_size"),
+            auth=kwargs.get("auth"),
+        )
+
+    async def query(self, database_id: str, **kwargs: Any) -> Any:
+        """Get a list of [Pages](https://developers.notion.com/reference/page) contained in the database.
+
+        *[🔗 Endpoint documentation](https://developers.notion.com/reference/post-database-query)*
+        """  # noqa: E501
+        return await self.parent.request(
+            path=f"databases/{database_id}/query",
+            method="POST",
+            query=pick(kwargs, "filter_properties"),
+            body=pick(kwargs, "filter", "sorts", "start_cursor", "page_size"),
+            auth=kwargs.get("auth"),
+        )
+
+    async def retrieve(self, database_id: str, **kwargs: Any) -> Any:
+        """Retrieve a [Database object](https://developers.notion.com/reference/database) using the ID specified.
+
+        *[🔗 Endpoint documentation](https://developers.notion.com/reference/retrieve-a-database)*
+        """  # noqa: E501
+        return await self.parent.request(
+            path=f"databases/{database_id}", method="GET", auth=kwargs.get("auth")
+        )
+
+    async def create(self, **kwargs: Any) -> Any:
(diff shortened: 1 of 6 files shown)


ramnes:
Hey @guacs! Thanks a lot for this first iteration.

Quick thoughts:
* I'd prefer to avoid code duplication, and I really want to avoid *manual* code duplication at all costs; if we don't find a simpler way, automated code duplication may be okay but I'd like to see a PoC first.
* I want us to stay as close as possible to the JS implementation, so we should:
  1. replicate the types defined [here](https://github.com/makenotion/notion-sdk-js/blob/main/src/api-endpoints.ts)
  2. expose them from `api_endpoints` (if we don't want to have a single big file, we could write a private `_types.py` module that we import in `api_endpoints`, and add add to its `__all__`)

Can't guarantee anything but I may try to work on a PoC as well. :)

In any case, please feel free to modify the code directly, that would actually help me to understand your ideas by providing me a diff to read!

guacs:
Yeah I definitely agree that the code duplication would definitely be a burden. I'll create a PoC for just the `UsersEndpoint` (this has the simplest types that would need to be implemented) with both an async version and a sync version created by `unasyncd`.

On the point of the types, I'll try my best to keep it in sync with the one that is in the typescript client, but there might be some changes that will need to be made to accommodate for the differences in the type system.

guacs:
@ramnes I have the basic setup with just the user responses typed. 

The `_sync_api_endpoints` was created using the `unasyncd` library I mentioned before. However, I did have to manually go and change the type of the client in the base endpoint class as well as rename it from `AsyncEndpoint` to `Endpoint` manually. The rest of it was automatically created with the following command: 
`unasyncd notion_client/_async_api_endpoints.py:notion_client/_sync_api_endpoints.py`

I ran the tests and all of them passed except for one test in helpers with the following error:

`tests/test_helpers.py::test_is_full_page_or_database - notion_client.errors.APIResponseError: API token is invalid.`

guacs:
Also, regarding the types I had some doubts:

- How closely do you want to follow the official TS types? Looking through it there is a lot of room for reusing already described types which are redefined in the TS types since that one is autogenerated.
- Do you prefer to keep the naming the same as that in the TS types or switch to more user-friendly type names? For example, the type of the response for the retrieve user endpoint is `UserObjectResponse` which could be simplified to `User` if we want to. Personally, I'm fine with either approach.

Also, please note that we will have to create some new types like `PersonDetails` due to the differences in how the types can be described in the two languages.

guacs:
Btw, mypy and pyright issues are for the following code:

```python
from notion_client.client import AsyncClient

async def main():

    async_client = AsyncClient()

    user = await async_client.users.retrieve("user-id")
```

ramnes:
What about something like this? Mypy seems happy:

```diff
 class UsersEndpoint(Endpoint):
-    def list(self, **kwargs: Any) -> SyncAsync[Any]:
+    @overload
+    async def list(self, **kwargs: Any) -> List[UserObjectResponse]:
+        ...
+
+    def list(self, **kwargs: Any) -> List[UserObjectResponse]:
         """Return a paginated list of [Users](https://developers.notion.com/reference/user) for the workspace.

         *[🔗 Endpoint documentation](https://developers.notion.com/reference/get-users)*
 ```

Or something like this, if we don't want to lie to type checkers:

```diff
 class UsersEndpoint(Endpoint):
-    def list(self, **kwargs: Any) -> SyncAsync[Any]:
+    @overload
+    def list(self, **kwargs: Any) -> Awaitable[List[UserObjectResponse]]:
+        ...
+
+    def list(self, **kwargs: Any) -> List[UserObjectResponse]:
         """Return a paginated list of [Users](https://developers.notion.com/reference/user) for the workspace.
```

guacs:
Mypy is not complaining for my code, but it is raising an error for the implementation within the library though. Do you have just one `overload`? I'm getting the following errors when running `mypy` on the `api_endpoints.py` file:

- `error: Single overload definition, multiple required  [misc]`
- `Overloaded function implementation cannot produce return type of signature 1  [misc]`

ramnes:
Yep right, not only it's not the correct way to use `@overload` since you always need multiple definitions with the decorator, but also a difference in return type is not enough since Mypy chooses the correct signature according to the parameters types. TIL. :)

Now this seems to work:

```diff
diff --git a/notion_client/api_endpoints.py b/notion_client/api_endpoints.py
index 8660dac..9a6c172 100644
--- a/notion_client/api_endpoints.py
+++ b/notion_client/api_endpoints.py
@@ -1,20 +1,22 @@
 """Notion API endpoints."""  # noqa: E501
-
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Awaitable, Dict, Generic, List, TypeVar, overload
 
 from notion_client.helpers import pick
 from notion_client.typing import SyncAsync
 
 if TYPE_CHECKING:  # pragma: no cover
-    from notion_client.client import BaseClient
+    from notion_client.client import AsyncClient, Client
+
+
+T = TypeVar("T", "AsyncClient", "Client")
 
 
-class Endpoint:
-    def __init__(self, parent: "BaseClient") -> None:
-        self.parent = parent
+class Endpoint(Generic[T]):
+    def __init__(self, parent: T) -> None:
+        self.parent: T = parent
 
 
-class BlocksChildrenEndpoint(Endpoint):
+class BlocksChildrenEndpoint(Endpoint[Any]):
     def append(self, block_id: str, **kwargs: Any) -> SyncAsync[Any]:
         """Create and append new children blocks to the block using the ID specified.
 
@@ -40,7 +42,7 @@ class BlocksChildrenEndpoint(Endpoint):
         )
 
 
-class BlocksEndpoint(Endpoint):
+class BlocksEndpoint(Endpoint[Any]):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
         self.children = BlocksChildrenEndpoint(*args, **kwargs)
@@ -109,7 +111,7 @@ class BlocksEndpoint(Endpoint):
         )
 
 
-class DatabasesEndpoint(Endpoint):
+class DatabasesEndpoint(Endpoint[Any]):
     def list(self, **kwargs: Any) -> SyncAsync[Any]:  # pragma: no cover
         """List all [Databases](https://developers.notion.com/reference/database) shared with the authenticated integration.
 
@@ -173,7 +175,7 @@ class DatabasesEndpoint(Endpoint):
         )
 
 
-class PagesPropertiesEndpoint(Endpoint):
+class PagesPropertiesEndpoint(Endpoint[Any]):
     def retrieve(self, page_id: str, property_id: str, **kwargs: Any) -> SyncAsync[Any]:
         """Retrieve a `property_item` object for a given `page_id` and `property_id`.
 
@@ -187,7 +189,7 @@ class PagesPropertiesEndpoint(Endpoint):
         )
 
 
-class PagesEndpoint(Endpoint):
+class PagesEndpoint(Endpoint[Any]):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
         self.properties = PagesPropertiesEndpoint(*args, **kwargs)
@@ -226,8 +228,16 @@ class PagesEndpoint(Endpoint):
         )
 
 
-class UsersEndpoint(Endpoint):
-    def list(self, **kwargs: Any) -> SyncAsync[Any]:
+class UsersEndpoint(Endpoint[Any]):
+    @overload
+    def list(self: "Endpoint[AsyncClient]", **kwargs: Any) -> Awaitable[List[Dict[str, Any]]]:
+        ...
+
+    @overload
+    def list(self: "Endpoint[Client]", **kwargs: Any) -> List[Dict[str, Any]]:
+        ...
+
+    def list(self, **kwargs: Any) -> Any:
         """Return a paginated list of [Users](https://developers.notion.com/reference/user) for the workspace.
 
         *[🔗 Endpoint documentation](https://developers.notion.com/reference/get-users)*
@@ -258,7 +268,7 @@ class UsersEndpoint(Endpoint):
         )
 
 
-class SearchEndpoint(Endpoint):
+class SearchEndpoint(Endpoint[Any]):
     def __call__(self, **kwargs: Any) -> SyncAsync[Any]:
         """Search all pages and child pages that are shared with the integration.
 
@@ -272,7 +282,7 @@ class SearchEndpoint(Endpoint):
         )
 
 
-class CommentsEndpoint(Endpoint):
+class CommentsEndpoint(Endpoint[Any]):
     def create(self, **kwargs: Any) -> SyncAsync[Any]:
         """Create a new comment in the specified page or existing discussion thread.
```

When used with this `users.py` file:

```python
import asyncio
import os
from pprint import pprint

from typing import TYPE_CHECKING

from notion_client import AsyncClient, Client


def main_sync() -> None:
    notion = Client(auth=os.environ["NOTION_TOKEN"])
    users = notion.users.list()

    if TYPE_CHECKING:
        reveal_type(users)

    pprint(users)


async def main_async() -> None:
    notion = AsyncClient(auth=os.environ["NOTION_TOKEN"])
    users = await notion.users.list()

    if TYPE_CHECKING:
        reveal_type(users)

    pprint(users)


main_sync()
asyncio.run(main_async())
```

It correctly reveals the types:

```
$ mypy users.py
users.py:15: note: Revealed type is "typing.Awaitable[builtins.list[builtins.dict[builtins.str, Any]]]"
users.py:25: note: Revealed type is "builtins.list[builtins.dict[builtins.str, Any]]"
Success: no issues found in 1 source file
```

Thoughts?

ramnes:
Yeah right, this is not the proper way to use `@overload`, it needs at least two alternatives and it can't use the return type to decide which alternative to use... TIL. :)

After a bit more digging, here is what I ended up with: https://github.com/ramnes/notion-sdk-py/compare/ramnes:1a8bdd4...ramnes:fe90378

With such a `users.py` file: 

```python
import asyncio
import os
from pprint import pprint

from typing import TYPE_CHECKING

from notion_client import AsyncClient, Client


def main_sync() -> None:
    notion = Client(auth=os.environ["NOTION_TOKEN"])
    users = notion.users.list()

    if TYPE_CHECKING:
        reveal_type(notion.users.parent)
        reveal_type(users)

    pprint(users)


async def main_async() -> None:
    notion = AsyncClient(auth=os.environ["NOTION_TOKEN"])
    users = await notion.users.list()

    if TYPE_CHECKING:
        reveal_type(notion.users.parent)
        reveal_type(users)

    pprint(users)


main_sync()
asyncio.run(main_async())
```

It correctly reveals the types:

```
$ mypy users.py
users.py:15: note: Revealed type is "notion_client.client.Client"
users.py:16: note: Revealed type is "builtins.list[builtins.dict[builtins.str, Any]]"
users.py:26: note: Revealed type is "notion_client.client.AsyncClient"
users.py:27: note: Revealed type is "builtins.list[builtins.dict[builtins.str, Any]]"
Success: no issues found in 1 source file
```

Thoughts?

Everything is in the [`poc-types`](https://github.com/ramnes/notion-sdk-py/tree/poc-types) branch if you want to play with it.

guacs:
This works perfectly! Yeah this is a far better approach. I think we can just go with this instead of the different implementations for sync and async.

guacs:
> Also, regarding the types I had some doubts:
> 
> * How closely do you want to follow the official TS types? Looking through it there is a lot of room for reusing already described types which are redefined in the TS types since that one is autogenerated.
> * Do you prefer to keep the naming the same as that in the TS types or switch to more user-friendly type names? For example, the type of the response for the retrieve user endpoint is `UserObjectResponse` which could be simplified to `User` if we want to. Personally, I'm fine with either approach.
> 
> Also, please note that we will have to create some new types like `PersonDetails` due to the differences in how the types can be described in the two languages.

Now I think we just need to make a decision regarding these points, and we can progress with the changes.

ramnes:
Cool, happy that you like it!

Just a quick note: I think we should use `__future__.annotations` to make a lot of the typing stuff simpler. It looks like it has been ported back to 3.7, which is great because ideally that's the first version I would like to support. If for some reason anything related to deferred annotations doesn't work in a particular version of Python, please tell me and we'll consider dropping support for it.

Regarding the types, I'd prefer that we stay consistent with the JavaScript SDK names. Thanks!

ramnes:
The PR is stale so I'll close it, but feel free to open a new PR if you feel like it!

PR #207, opened 2023-09-15
Adding "after"-parameter to children endpoint

timo-codecentric:
To insert a block in a specific place in a page, the "after" parameter is necessary, which was missing in the API call and is added by this PR. See https://developers.notion.com/reference/patch-block-children


ramnes:
Good catch, thanks! CI fails but that seems unrelated, will check that later.

brainyoung-com:
It's confusing to insert a block before the first block of the page.
Is it possible? Or does it require any trick? thanks.

PR #208, opened 2023-10-02
Fix BlockChildrenEndpoint not exposed via Client interface

CaptureTheFrog:
Block children endpoint was not accessible via client interface so unusable despite seemingly being implemented otherwise


ramnes:
It's already available under `notion.blocks.children`, just like the JS SDK. I would likely merge a PR that improves the docs for this.

PR #211, opened 2023-11-10
Fix dead link in examples/README.md

cbrenton:
Changes the link from `docs/quick_start.md` to `docs/user_guides/quick_start.md`


ramnes:
Good catch, thanks!

PR #220, opened 2023-12-22
Make sure nullable attributes icon and cover are set to None

### LINKED ISSUE #219: Deleting an icon or a cover is not possible.
Analogue to the discussion in #128, it's currently not possible to use notion-sdk-py to delete an icon or a cover. The problem is in  `api_endpoints.py` and the way how `pick` just ignores args set to `None`.

FlorianWilhelm:
This fixes #219.

It would be really nice if this could be merged and released so that I can also make a new release of [Ultimate Notion](https://ultimate-notion.com/). Thank you 🙏

### DIFF
--- a/notion_client/api_endpoints.py
+++ b/notion_client/api_endpoints.py
@@ -226,10 +226,15 @@ class PagesEndpoint(Endpoint):
 
         *[🔗 Endpoint documentation](https://developers.notion.com/reference/patch-page)*
         """  # noqa: E501
+        body = pick(kwargs, "archived", "properties", "icon", "cover")
+        for nullable_attr in ("icon", "cover"): # make sure those are set to `None`
+            if kwargs.get(nullable_attr, object) is None:
+                body[nullable_attr] = None
+                
         return self.parent.request(
             path=f"pages/{page_id}",
             method="PATCH",
-            body=pick(kwargs, "archived", "properties", "icon", "cover"),
+            body=body,
             auth=kwargs.get("auth"),
         )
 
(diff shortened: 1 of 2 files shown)


ramnes:
Thanks for the PR @FlorianWilhelm! I went with a bit more extreme approach in 28020cd, can you tell me if that works for  you? See the rationale in #101.

ramnes:
PS – Regarding Ultimate Notion, I'm always glad to see some effort to build a higher-level library on top of notion-sdk-py so thank you for this! Did you check [notional](https://github.com/jheddings/notional)? It's been there for a while now and seems to be willing to achieve the same goals as you are and would probably benefit from more contributors. :)

FlorianWilhelm:
@ramnes Thanks for fixing this issue and I think your approach is even nicer than my PR. 

Regarding [notional](https://github.com/jheddings/notional), I tried to use it, but it doesn't seem to be under active development anymore, and I don't agree with many of the design decision. This is why I copied some parts over as I like the Pydantic approach, but added an extra layer of indirection, which famously solves almost all problems in computer science 😅. I also think that [Ultimate Notion](https://ultimate-notion.com/) is much easier to use compared to Notional and of course more *pythonic*. But doesn't every developer say that about his/her own software? ;-) 

I'll definitely contribute to notion-sdk-py whenever I find an issue, as your package is the base for my package. Sadly, I also found a lot of bugs within the Notion API itself, but the Notion devs are really not that responsive. For instance, whenever I try to make a self-referential two-way relation, for instance to express a parent task and child tasks in a task database, the Notion API only returns a one-way relation. Creating a two-way relation from a database A to a database B works perfectly but only if A is not B.
What are your experiences with the Notion devs and bugs in their API or is it just me 😅 ?

ramnes:
Got to admit that I haven't talked to them much so far. Did you try the Notion API developers Slack?

PR #222, opened 2023-12-24
Allow passing starting cursor in paginated helper

smarttommyau:
Allow passing next_cursor to the paginated api to set the starting cursor.


ramnes:
I've pushed a commit to stay closer to the TS SDK (and fix CI), does it work for you?

ramnes:
Going ahead and merging, please comment if anything's wrong. :)

PR #229, opened 2024-01-09
Switch to Ruff

baydinvladislav:


ramnes:
Cool, let's fix CI and remove the redundant hooks then!

ramnes:
Nice, thanks!

PR #250, opened 2024-12-30
Update to vcrpy-6.0.2

ramnes:


ramnes:
Updated, seems to work now. Not sure what happened here!

Issue #260, opened 2025-03-14
Add OAuth revoke and introspect endpoints


ramnes:
OAuth revoke and introspect endpoints got added to the JS SDK: https://github.com/makenotion/notion-sdk-js/pull/552

Let's add them here as well. :)

vincedbowen:
I'd love to work on this! Are there any additional requirements I should be aware of?

ramnes:
Go ahead!

Only requirements is for the DX / library API to be identical to the JS SDK, and for our Python code to be consistent with what we already have. :)

vincedbowen:
Great, thanks :-) It looks like the tests I would write on this would be dependent on the access token's being generated. Is there anyway to generate these without a public integration?

ramnes:
I don't think so, but we only need to burn one token to register the VCR.py cassette. Subsequent tests won't revoke the token as long as this cassette is being used.

vincedbowen:
Sorry if I am not understanding 😅 ! From my understanding of the Read me "To create new tests or run them without cassettes, you need to set up the environment variables `NOTION_TOKEN` and `NOTION_TEST_PAGE_ID`", this means I would have to generate my own token to develop any additional tests. The pytest-vcr docs are a bit minimal, so I am not totally sure how I would add new tests in the cassette.

ramnes:
Yes, you need to create an integration so that you can generate a new token manually and burn it while generating the cassette for this test.

vincedbowen:
Perfect! Thanks for the explanation

vincedbowen:
I have the endpoints working with tests for each endpoint, using the method of burning a token to generate the cassette for the test. I did notice that in the `client.py`, the authorization is in the form `Bearer {auth}`, but from the docs, the authorization looks to be `Basic '"$BASE64_ENCODED_ID_AND_SECRET"'`. Will this be a problem for users?

ramnes:
Oh, good catch! It looks like we missed this commit from two years ago: https://github.com/makenotion/notion-sdk-js/commit/0877d3478c680bf6b5860cfc21e7ab012084aad4

We should implement this "Get token" endpoint as well, and add the same changes for authorization, i.e. access either the token or the client ID and secret combo, which should be only usable for the OAuth-related endpoints, as in the JS SDK.

We can handle this in a separate issue and PR, or address everything at once, whichever you prefer.

ramnes:
[...]

PR #261, opened 2025-03-15
Adding OAuth revoke and introspect endpoints

vincedbowen:
As described in Issue #260, 

> OAuth revoke and introspect endpoints got added to the JS SDK: https://github.com/makenotion/notion-sdk-js/pull/552

This PR adds that functionality to this SDK.

### DIFF
--- a/notion_client/api_endpoints.py
+++ b/notion_client/api_endpoints.py
@@ -325,3 +325,29 @@ class CommentsEndpoint(Endpoint):
             query=pick(kwargs, "block_id", "start_cursor", "page_size"),
             auth=kwargs.get("auth"),
         )
+
+
+class OAuthEndpoint(Endpoint):
+    def introspect(self, **kwargs: Any) -> SyncAsync[Any]:
+        """Introspect the provided token.
+
+        *[🔗 Endpoint documentation](https://developers.notion.com/reference/introspect-token)*
+        """
+        return self.parent.request(
+            path="oauth/introspect",
+            method="POST",
+            body=pick(kwargs, "token"),
+            auth=kwargs.get("auth"),
+        )
+
+    def revoke(self, **kwargs: Any) -> SyncAsync[Any]:
+        """Revoke the provided token.
+
+        *[🔗 Endpoint documentation](https://developers.notion.com/reference/revoke-token)*
+        """
+        return self.parent.request(
+            path="oauth/revoke",
+            method="POST",
+            body=pick(kwargs, "token"),
+            auth=kwargs.get("auth"),
+        )
--- a/notion_client/client.py
+++ b/notion_client/client.py
@@ -16,6 +16,7 @@ from notion_client.api_endpoints import (
     PagesEndpoint,
     SearchEndpoint,
     UsersEndpoint,
+    OAuthEndpoint,
 )
 from notion_client.errors import (
     APIResponseError,
@@ -77,6 +78,7 @@ class BaseClient:
         self.pages = PagesEndpoint(self)
         self.search = SearchEndpoint(self)
         self.comments = CommentsEndpoint(self)
+        self.oauth = OAuthEndpoint(self)
 
     @property
     def client(self) -> Union[httpx.Client, httpx.AsyncClient]:
--- a/tests/cassettes/test_introspect_token.yaml
+++ b/tests/cassettes/test_introspect_token.yaml
@@ -1,0 +1,29 @@
+interactions:
+- request:
+    body: '{"token": "ntn_..."}'
+    headers:
+      accept:
+      - '*/*'
+      accept-encoding:
+      - gzip, deflate
+      authorization:
+      - Basic '"$BASE64_ENCODED_ID_AND_SECRET"'
+      connection:
+      - keep-alive
+      content-length:
+      - '63'
+      content-type:
+      - application/json
+      host:
+      - api.notion.com
+      notion-version:
+      - '2022-06-28'
+    method: POST
+    uri: https://api.notion.com/v1/oauth/introspect
+  response:
+    content: '{"active":true,"scope":"read_content insert_content update_content read_user_with_email
+      read_user_without_email","iat":1742002165921,"request_id":"14314d51-f34c-47a0-886b-4d792d9dad0c"}'
+    headers: {}
+    http_version: HTTP/1.1
+    status_code: 200
+version: 1
--- a/tests/cassettes/test_revoke_token.yaml
+++ b/tests/cassettes/test_revoke_token.yaml
@@ -1,0 +1,28 @@
+interactions:
+- request:
+    body: '{"token": "ntn..."}'
+    headers:
+      accept:
+      - '*/*'
+      accept-encoding:
+      - gzip, deflate
+      authorization:
+      - Basic '"$BASE64_ENCODED_ID_AND_SECRET"'
+      connection:
+      - keep-alive
+      content-length:
+      - '63'
+      content-type:
+      - application/json
+      host:
+      - api.notion.com
+      notion-version:
+      - '2022-06-28'
+    method: POST
+    uri: https://api.notion.com/v1/oauth/revoke
+  response:
+    content: '{"request_id":"ba61313e-c752-49bc-858d-0d961eda422e"}'
+    headers: {}
+    http_version: HTTP/1.1
+    status_code: 200
+version: 1
--- a/tests/test_endpoints.py
+++ b/tests/test_endpoints.py
@@ -199,3 +199,15 @@ def test_pages_delete(client, page_id):
     assert response
 
     client.pages.update(page_id=page_id, archived=False)
+
+
+@pytest.mark.vcr()
+def test_revoke_token(client, token):
+    response = client.oauth.revoke(token=token)
+    assert response
+
+
+@pytest.mark.vcr()
+def test_introspect_token(client, token):
+    response = client.oauth.introspect(token=token)
+    assert response


notion_client/api_endpoints.py
```diff
@@ -325,3 +325,29 @@ def list(self, **kwargs: Any) -> SyncAsync[Any]:
             query=pick(kwargs, "block_id", "start_cursor", "page_size"),
             auth=kwargs.get("auth"),
         )
+
+
+class OAuthEndpoint(Endpoint):
+    def introspect(self, **kwargs: Any) -> SyncAsync[Any]:
+        """Introspect the provided token.
```
ramnes: Nit: Let's stay close to the Notion docs wording.

```suggestion
        """Get the token's active status, scope, and issued time.
```

notion_client/api_endpoints.py
```diff
@@ -325,3 +325,29 @@ def list(self, **kwargs: Any) -> SyncAsync[Any]:
...
+
+        *[🔗 Endpoint documentation](https://developers.notion.com/reference/introspect-token)*
+        """
```
ramnes:
```suggestion
        """  # noqa: E501
```

notion_client/api_endpoints.py
```diff
@@ -325,3 +325,29 @@ def list(self, **kwargs: Any) -> SyncAsync[Any]:
...
+            method="POST",
+            body=pick(kwargs, "token"),
+            auth=kwargs.get("auth"),
+        )
+
+    def revoke(self, **kwargs: Any) -> SyncAsync[Any]:
+        """Revoke the provided token.
+
+        *[🔗 Endpoint documentation](https://developers.notion.com/reference/revoke-token)*
+        """
```
ramnes:
```suggestion
        """  # noqa: E501
```

tests/cassettes/test_introspect_token.yaml
```diff
@@ -0,0 +1,29 @@
+interactions:
+- request:
+    body: '{"token": "ntn_..."}'
+    headers:
+      accept:
+      - '*/*'
+      accept-encoding:
+      - gzip, deflate
+      authorization:
+      - Basic '"$BASE64_ENCODED_ID_AND_SECRET"'
```
ramnes: Hmm, how is it possible that the cassette contains this Authorization header if we didn't implement it? Do the endpoint also work with a normal token, and then you changed the cassette manually? (We should never modify the cassettes manually.)

ramnes (commented):
LGTM overall, just need some clarification on the Authorization header.

notion_client/api_endpoints.py
ramnes: Hmm, curious that the CI doesn't break without these. Maybe we can remove them all for some reason?

tests/cassettes/test_introspect_token.yaml
vincedbowen: I did modify the cassette 😅. When I ran the test originally, I hard-coded the authorization because of the difference mentioned in #260. Even with those changes fixed, won't the authorization and token not be obfuscated in the cassette, or if set correctly they will?

notion_client/api_endpoints.py
vincedbowen: I can add them to be consistent with the rest of the SDK if that would be best.

tests/cassettes/test_introspect_token.yaml
ramnes: I think you just need to update VCR config accordingly: https://github.com/ramnes/notion-sdk-py/blob/b9d8d83e9333ba98d6f2124b13a364e1cbdd5a9f/tests/conftest.py#L11-L25

You probably want to update the `before_record_response` hook so that it masks the returned token as well, in the body (I guess you updated that one as well!) :)

vincedbowen:
I think some documentation updates would be great, but I will wait for any feedback on codebase changes before doing that so I don't waste anytime documenting unwanted changes :-)! The tests are a bit annoying since the developer effectively has to burn two tokens, but that was the only strategy I found to work. Let me know if any changes should be made!!

notion_client/client.py
```diff
@@ -44,7 +48,7 @@ class ClientOptions:
         notion_version: Notion version to use.
     """
 
-    auth: Optional[str] = None
+    auth: Optional[Union[str, tuple[str, str]]] = None
```
ramnes: The JS SDK doesn't support this kind of auth at this level AFAICT, so this should be reverted.

notion_client/client.py
```diff
@@ -93,7 +98,15 @@ def client(self, client: Union[httpx.Client, httpx.AsyncClient]) -> None:
...
-            client.headers["Authorization"] = f"Bearer {self.options.auth}"
+            if isinstance(self.options.auth, tuple):
+                client_id = self.options.auth[0]
+                client_secret = self.options.auth[1]
+                auth_header = base64.b64encode(
+                    f"{client_id}:{client_secret}".encode()
+                ).decode("utf-8")
+                client.headers["Authorization"] = f'Basic "{auth_header}"'
+            else:
+                client.headers["Authorization"] = f"Bearer {self.options.auth}"
```
ramnes: Same thing here, it looks like the JS SDK doesn't support this kind of auth at this level and this should be reverted.

notion_client/client.py
```diff
@@ -102,11 +115,19 @@ def _build_request(
         path: str,
         query: Optional[Dict[Any, Any]] = None,
         body: Optional[Dict[Any, Any]] = None,
-        auth: Optional[str] = None,
+        auth: Optional[Union[str, tuple[str, str]]] = None,
```
ramnes: Feels like a TypedDict would be more appropriate than a tuple here (and closer to the JS SDK implementation), what do you think?

notion_client/client.py
```diff
@@ -120,6 +141,11 @@ def _parse_response(self, response: Response) -> Any:
             try:
                 body = error.response.json()
                 code = body.get("code")
+                # Any oauth errors throw this exact error syntax, so handle them as so
+                if "code" not in body and body.get("error") == "invalid_client":
+                    raise APIResponseError(
+                        response, body["error"], APIErrorCode("unauthorized")
+                    )
```
ramnes: This is weird, how does the JS SDK handle this?

tests/conftest.py
```diff
@@ -14,13 +15,45 @@ def remove_headers(response: dict):
...
+
+            except json.JSONDecodeError:
+                pass
+
+        return response
+
     return {
         "filter_headers": [
-            ("authorization", "ntn_..."),
+            ("authorization", "ntn_... OR base64_encoded(client_id:client_secret)"),
```
ramnes: Can't we use `before_record_request` to actually record the correct authorization type?

ramnes (commented):
See comments.

tests/conftest.py
```diff
@@ -14,13 +15,45 @@ def remove_headers(response: dict):
...
+                if "token" in body_json:
+                    body_json["token"] = "ntn_..."
+                if "code" in body_json:
+                    body_json["code"] = "..."
+                if "redirect_uri" in body_json:
+                    body_json["redirect_uri"] = "http://..."
+                request.body = json.dumps(body_json).encode("utf-8")
+
+            except (json.JSONDecodeError, AttributeError):
+                pass
```
ramnes: Why do we need to silence the exception here? Is it expected that we have a JSON decode error or attribute error and do not care at all?

tests/conftest.py
```diff
@@ -14,13 +15,45 @@ def remove_headers(response: dict):
...
+        if "content" in response:
+            try:
+                content_json = json.loads(response["content"])
+                if "access_token" in content_json:
+                    response["content"] = json.dumps(
+                        {key: "..." for key in content_json}, separators=(",", ":")
+                    )
+
+            except json.JSONDecodeError:
+                pass
```
ramnes: Same question here: why do we need to silence the exception? Is it expected that we get a JSON decode error here and do not care?

notion_client/client.py
```diff
@@ -102,11 +107,19 @@ def _build_request(
         path: str,
         query: Optional[Dict[Any, Any]] = None,
         body: Optional[Dict[Any, Any]] = None,
-        auth: Optional[str] = None,
+        auth: Optional[Union[str, Dict[str, str]]] = None,
```
ramnes: Let's use a [TypedDict](https://peps.python.org/pep-0589/) :P 

(and drop support for Python 3.7 since it has been implemented in 3.8)

tests/conftest.py
```diff
@@ -14,13 +15,52 @@ def remove_headers(response: dict):
...
+
+        return response
+
+    def scrub_auth_header(key: str, value: str, request: Optional[object] | None):
```
ramnes: I think the type hint for `request` is wrong here, and likely to be the source of the CI errors: `Optional` already means "something or `None`", so this union probably doesn't make sense, and IIRC unions require the `Union` type for older Python versions.

Also I'm quite certain we can make this `object` type more precise. :)

vincedbowen:
Let me know what you think of the current changes since your last comments. If those look better, I will fix the remaining changes, and make any documentation updates :-).

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py37,py38,py39,py310,py311,py312,py313
+envlist = py38,py39,py310,py311,py312,py313
```
ramnes, line 2: Also needs to be updated in `README.md`, `setup.py`, and `.github/workflows/test.yml`

ramnes:
> Let me know what you think of the current changes since your last comments. If those look better, I will fix the rest of the pending changes :-).

Yeah it's going in the right direction, keep up the good work! You've already done more than I thought necessary when I opened #261. Sorry if the PR feels a bit tedious, this library is vastly used in the wild so the last thing I want is more work fixing things down the line. :)

vincedbowen:
> > Let me know what you think of the current changes since your last comments. If those look better, I will fix the rest of the pending changes :-).
> 
> Yeah it's going in the right direction, keep up the good work! You've already done more than I thought necessary when I opened #261. Sorry if the PR feels a bit tedious, this library is vastly used in the wild so the last thing I want is more work fixing things down the line. :)

No worries, apologies if it has taken a number of iterations to produce high quality code! 

I think I have resolved all of your reviews, besides the error handling in the case of the 401 response. To Clarify why I am confused on this:

If I do nothing, an APIResponseError is not raised, and the error is instead raised as an HTTPResponseError. This is because the current code checks for the string `code`, and matches that to a number status. An example of the response body where this works just fine is [here](https://developers.notion.com/reference/retrieve-a-comment) (if you click the 403 example response). The authentication APIs do not respond with the `code` in the response body, and thus the error cannot be matched. The JS SDK only tests for a successful response as seen [here](https://github.com/makenotion/notion-sdk-js/blob/e623fc4f7ba3a5a054f30cd9e9132f096c078cdc/test/Client.test.ts), so this error never needs to be handled. I see three ways to solve this issue:

1. Let the error be passed to the HTTPResponseError (This seems a bit hacky, as this is an APIResponseError where Notion is telling the user they are unauthorized)
2. Use the actual response code from httpx instead of accessing the code through the response body
3. Raise an issue or conversation with the folks who run the JS SDK about why the response errors are handled in this way

Let me know what you think of those options, or if you have an entirely different idea on how to solve that :-). 
 
Additionally here are the steps I have to take in order to create new cassettes for the tests. This will explain why I have to generate two token. I don't think there is a way to avoid this because the token would have to be generated on the fly and the environment variable would have to be set from within the tests. I don't think that is a solid solution because the tests shouldn't have to be executed in a definitive order like that.
1. I created a public integration
2. I set the environment variables `NOTION_CLIENT_ID`, `NOTION_CLIENT_SECRET`, and `NOTION_REDIRECT_URI`
3. I followed the authorization URL in my public integration (https://api.notion.com/v1/oauth/authorize?client_id=<YOUR_CLIENT_ID>&response_type=code&owner=user&redirect_uri=<YOUR_REDIRECT_URI>)
  i. I selected the pages and allowed access for my account
  ii. I grabbed the code from the redirect URI http://localhost:3000/callback?code=<YOUR_CODE>&state=
4. I set this as the `NOTION_CODE`. This is the code that will be used to actually generate a token in the `test_token(client, redirect_uri, code, client_id, client_secret)` test.
5. Because of the aforementioned reasons I followed step 3 again to get a new code
6. I used a cURL command to generate a token from this code. This token is what will be used for the introspect and revoke tests. 
```
curl --location 'https://api.notion.com/v1/oauth/token' \
--header 'Content-Type: application/json' \
--header 'Notion-Version: 2022-06-28' \
--header 'Authorization: Basic "<YOUR_Base64Encoded($client_id:$client_secret)>"' \
--data '{
  "grant_type": "authorization_code",
  "code": "<YOUR_CODE>",
  "redirect_uri": "<YOUR_REDIRECT_URI>"
}'
```
7. I grabbed the "access_token" from the response body and set that as the `NOTION_TOKEN` environment variable

I understand this is a bit of a clunky and hacky solution, but it is the best I could think of right now. If you have a better way to go about this, please let me know! Otherwise I will update the developer docs to include these steps. 

Apologies for the wordy comment 😅

tests/conftest.py
```diff
@@ -14,13 +16,64 @@ def remove_headers(response: dict):
...
+                    body_json["code"] = "..."
+                if "redirect_uri" in body_json:
+                    body_json["redirect_uri"] = "http://..."
+                request.body = json.dumps(body_json).encode("utf-8")
+            except json.JSONDecodeError as e:
+                raise json.JSONDecodeError(
+                    f"Failed to decode request body: {request.body} \n Error occurred at {e.pos} with message: {e.msg}",
+                    request.body,
+                    e.pos,
+                )
```
ramnes: Why not just let it raise? What's the point of the try-except construct here?

tests/conftest.py
```diff
@@ -14,13 +16,64 @@ def remove_headers(response: dict):
...
+        return request
+
+    def scrub_response(response: dict):
+        if "content" in response:
+            content = response["content"]
+            # Like the case tests/cassettes/test_api_async_request_bad_request_error.yaml, where the response is just a string, not JSON
+            # We don't want to raise an error here because the response is not JSON and that is ok
```
ramnes, line 36: Don't we have the Content-Encoding header for this?

Edit: I mean, before we remove it! :D

tests/conftest.py
```diff
@@ -14,13 +16,64 @@ def remove_headers(response: dict):
         response["headers"] = {}
         return response
 
+    def scrub_requests(request: dict):
```
ramnes: Is `request` really a `dict`? I don't think you can do `request.body` on a Python dict. 👀

vincedbowen:
> But does the JS SDK return an HTTPResponseError or an APIResponseError?
>
>    If it returns an HTTPResponseError, we should do 1., merge this PR, and then do 3. :)
>    If it returns an APIResponseError, we should replicate the way it detects the message as an APIResponseError.

From my understanding it actually returns an UnkownHTTPResponseError. If I spin up a quick testing JS file and call the introspect function with invalid credentials, the response returns this `UnknownHTTPResponseError: Request to Notion API failed with status: 401`. So I have removed that check for now, and am just returning an HTTPResponseError because there is no Unknown version of this in this SDK.

vincedbowen:
> Thanks for the detailed explanation. I didn't realize that the authorization flow required a browser, that's really cumbersome for our tests. What would you think of using [pytest-mock](https://github.com/pytest-dev/pytest-mock/) for test_token and test_revoke_token rather than cassettes?

That would work for me!! I will set that up now!

notion_client/api_endpoints.py
```diff
@@ -325,3 +325,41 @@ def list(self, **kwargs: Any) -> SyncAsync[Any]:
...
+class OAuthEndpoint(Endpoint):
+    def token(self, **kwargs: Any) -> SyncAsync[Any]:
+        """Creates an access token that a third-party service can use to authenticate with Notion.
+
+        *[🔗 Endpoint documentation](https://developers.notion.com/reference/create-a-token)*
+        """  # noqa: E501
+        return self.parent.request(
+            path="oauth/token",
+            method="POST",
+            body=pick(kwargs, "grant_type", "code", "redirect_uri"),
```
ramnes, line 339: PR 595 on notion-sdk-js added more to it: https://github.com/makenotion/notion-sdk-js/pull/595

PR #264, opened 2025-04-16
Add more explicit example for page creation

seanmann-synth:
Just updating the example to be more clear how to create a page with content as I was a little confused on the necessary format and had to dig a bit to find clarity.


seanmann-synth:
Looks like the Code Quality / check is deprecated as of yesterday 👎

ramnes:
Thanks!

Issue #268, opened 2025-05-21
Add file upload endpoints


ramnes:
We need to port this PR: https://github.com/makenotion/notion-sdk-js/pull/565

ramnes:
This related PR as well: https://github.com/makenotion/notion-sdk-js/pull/566

PR #270, opened 2025-06-23
Support file uploads

tsinglinrain:
# Overview
This PR adds support for file uploads endpoints to the notion-sdk-py, following the design and conventions of the official [Notion API documentation](https://developers.notion.com/reference/create-a-file-upload) and the [notion-sdk-js](https://github.com/makenotion/notion-sdk-js). https://github.com/ramnes/notion-sdk-py/issues/268

# Implementation Details
`notion_client/api_endpoints.py`
Following the existing coding style, I referred to the official documentation and the file uploads implementation in notion-sdk-js when designing the endpoint logic.
The request uses files to handle binary content and data for other metadata, keeping them clearly separated.

`notion_client/client.py`
I extended the existing request logic to support files and data as additional parameters. The code now dynamically constructs the request based on parameter type.

`notion_client/helper.py`
The official JavaScript SDK only supports single-chunk uploads and relies on external libraries for splitting large files. In Python, there is no audited or popular native implementation, so I wrote a custom `split_file` function.

`split_file`: synchronously splits files into smaller chunks.

`split_file_async`: provides similar functionality using asynchronous iteration.

These utility functions are designed to be reusable and thus included in helper.py.

`notion_client/notion_mime_detector.py`
Due to inconsistencies in Python’s built-in mimetypes across platforms (e.g., Windows 11 registry values override expected MIME types), I implemented a dedicated MIME type detector.
This module ensures platform-consistent behavior and explicitly supports the MIME types Notion accepts. It also includes a utility for listing supported file types.

`examples/file_uploads`
Includes usage examples to demonstrate the new file upload functionality.

# Additional Thoughts
In the future, this feature could be further abstracted into a higher-level interface where the user only needs to provide the file path and name, and the SDK handles the rest (e.g., auto-selecting upload mode).

# Final Note
- I’ve been a long-time user of [notion-sdk-py](https://github.com/ramnes/notion-sdk-py) and truly appreciate the work done by the maintainers.

- This is my first PR with a relatively large amount of code, and I’m still learning. There may be issues or non-idiomatic patterns in the implementation, and I warmly welcome any suggestions or corrections. Thank you for your time and guidance!


FlorianWilhelm:
Hi @tsinglinrain, cool that you started to work on this feature. May I ask what the current status is? 
When this PR is merged, I am hoping to implement the same feature in [Ultimate Notion](https://github.com/ultimate-notion/ultimate-notion), which relies on notion-sdk-py. So the sooner you are ready, the sooner I can start working on a high-level implementation 😄

tsinglinrain:
Hi, @FlorianWilhelm, sorry for the delay. The last commit already completed all the intended work. I had originally planned to do another round of checks for any potential issues, but I got caught up with job hunting and the related preparations, and couldn't find the time.

I'll be taking two weeks off starting tomorrow, and during that time I’ll be fully dedicating myself to finishing up this PR.

> Hi @tsinglinrain, cool that you started to work on this feature. May I ask what the current status is? When this PR is merged, I am hoping to implement the same feature in [Ultimate Notion](https://github.com/ultimate-notion/ultimate-notion), which relies on notion-sdk-py. So the sooner you are ready, the sooner I can start working on a high-level implementation 😄

tsinglinrain:
Hi, when you have a moment, could you please review this PR? I believe the file upload feature is mostly done in the latest commit. There might still be some issues, and I plan to resolve them within the next two weeks.

notion_client/client.py
```diff
@@ -102,42 +106,104 @@ def _build_request(
...
+                headers=headers,
+            )
+        else:
+            return self.client.build_request(
+                method,
+                path,
+                params=query,
+                json=body,
+                headers=headers,
+            )
```
ramnes: `key` can only be `"file"` here, which is aligned with the Notion docs [saying that "file" must be set](https://developers.notion.com/docs/uploading-small-files#step-2-upload-file-contents), but as far as I can see in the Notion TS SDK, nothing prevents their users to set other keys as files. They actually match on the content: https://github.com/makenotion/notion-sdk-js/blob/f84957479f2b8d6f17b52972879f3c330c8ae7c0/src/Client.ts#L234-L246

I think we shouldn't make assumptions and just have a similar logic than the TS SDK, so that we can write things like this (example pythonic API inspired from HTTPX, since we obviously can't use TS `FormData` here):

```python
form_data_params = {
    "avatar": ("pic.jpg", open("pic.jpg", "rb"), "image/jpeg"),
    "resume": ("cv.pdf", open("cv.pdf", "rb"), "application/pdf"),
    "name": "Guillaume"
}
```

Eventually the API will say that it only expects `"file"` but that does not seem to be the client responsibility according to the TS SDK.

So all in all, I'd suggest something like this (also, let's use an early return here):

```python
if not form_data:
    return self.client.build_request(...)

files = {}
data = {}
for key, value in form_data_params.items():
    if isinstance(value, tuple) and len(value) >= 2:
        # Assumed to be (filename, fileobj, [content_type])
        files[key] = value
    elif hasattr(value, "read"):
        # Raw file-like object without filename/content-type
        files[key] = ("upload", value)
    else:
        data[key] = value

self.client.build_request(...)
```

If you like the idea, we can even add our own wrapper:

```python
from dataclasses import dataclass
from typing import IO

@dataclass
class UploadFile:
    filename: str
    fileobj: IO
    content_type: str = "application/octet-stream"

 ...
    if isinstance(value, UploadFile):
        files[key] = (value.filename, value.fileobj, value.content_type)
 ```
 
 Or a `FormData` class that mimics TS one.

File: notion_client/client.py [hunk 2]
```diff
@@ -102,42 +106,104 @@ def _build_request(
...
+                    self.logger.debug(
+                        f"Response content encoding: {response.headers.get('content-encoding', 'none')}"
+                    )
+                    self.logger.debug(f"Content starts with: {content[:20].hex()}")
+                    raise e
+            else:
+                self.logger.error(f"JSON decode error: {e}")
+                self.logger.debug(f"Response content: {content[:200]!r}")
+                raise e
+
```
ramnes: What is this diff and how is it related to file uploads? Can we do something simpler?

ramnes: HTTPX is supposed to automatically decompress gzip: https://www.python-httpx.org/quickstart/#binary-response-content

tests/conftest.py
```diff
@@ -8,10 +8,84 @@
 from notion_client import AsyncClient, Client
 
 
+# Fix for VCR compatibility issue with cassettes that have gzip content
```
ramnes, line 11: What issue? :)

ramnes (changes requested):
Thanks for the work! C.f. comments for the change requests.

I may be wrong but I feel like we should be able to simplify things related to gzip, the code here seems unnecessarily low-level. (Notion SDK codebase [doesn't even contain "gzip"!](https://github.com/search?q=repo%3Amakenotion%2Fnotion-sdk-js%20gzip&type=code))

notion_client/client.py
tsinglinrain: I agree with your point and I fully share your view on that.

Also, I have thoughts on other aspects:

1. I find using `UploadFile` a bit cumbersome, so I didn’t add it. If users need to import this class, it would require `notion_client\__init__.py` to additionally export it.
2. In fact, the TS code also includes a type check for `string`, and the Notion documentation specifies the type as well ([Send a file upload](https://developers.notion.com/reference/send-a-file-upload)). That’s why I added a type check here.
3. Changing it to `files[key] = value` allows `httpx` to automatically determine the MIME type, making the upload process simpler.

```python
if not form_data:
    return self.client.build_request(...)

files: Dict[str, Any] = {}
data: Dict[str, Any] = {}
for key, value in form_data.items():
    if isinstance(value, tuple) and len(value) >= 2:
        files[key] = value
    elif hasattr(value, "read"):
        files[key] = value
    elif isinstance(value, str):
        data[key] = value
    else:
        data[key] = str(value)

return self.client.build_request(...)
```

notion_client/client.py
tsinglinrain: This part was added earlier due to the gzip issue, which required some additional handling here. It has now been completely removed.

tests/conftest.py
tsinglinrain: While running the tests, I encountered an `AttributeError: 'NoneType' object has no attribute 'encode'`. To properly handle cases where `content` is `None`, I added this function.

This piece of code was written with the help of AI, as I’m not very familiar with testing. That’s why I didn’t move the third-party import statements to the top, but instead directly added the function, making changes while preserving as much of the original code as possible for easier review. I‘m sorry, and I’m afraid I may not be able to provide detailed explanations or make further modifications to this part.

tsinglinrain:
Could you please review this again? I've already written replies.@ramnes 
By the way, is it necessary to require `vcrpy < 6`? I tested with `vcrpy==6.0.2` as well as the latest version (which does not support Python 3.8), and using the main branch version — without any changes to `tests/conftest.py` — my local tests ran without errors.

FlorianWilhelm:
Hi @tsinglinrain, thank you. I think it never hurts to kindly ping the corresponding maintainer, that is @ramnes, so that he gets a small notification by Github 😃

examples/file_uploads/create_file_uploads_extenal.py
```diff
@@ -0,0 +1,93 @@
...
+    Wait for file upload/import to complete.
+
+    Args:
+        file_upload_id: The file upload ID.
+        poll_interval: Polling interval in seconds.
+        max_wait_time: Maximum wait time in seconds.
+    """
+    start_time = time.time()
+
+    while time.time() - start_time < max_wait_time:
```
adamtheturtle: I am not a maintainer of this project, just a user.

That said, I recommend you consider using [`time.monotonic`](https://docs.python.org/3/library/time.html#time.monotonic).

`time.time()` can jump forwards or go backwards.

tsinglinrain: Thanks for pointing this out, and I’ve revised the code to use `time.monotonic`.

ramnes:
Sorry for the delay here. I'm not entirely sure of everything I see in the diff, but the core changes seems fine so I'll go ahead, merge, and open follow up issues. Thanks a lot for your work and patience, much appreciated! And as @FlorianWilhelm said, feel free to ping me when needed.

ramnes:
Sure! Just released 2.5.0.

PR #281, opened 2025-09-01
added "attachements" and "display_name" to body

zubu007:
Added the following parameters in the body for the POST request in `CommentsEndpoint` class:

* attachments
* display_name


ramnes:
Thanks!

PR #282, opened 2025-09-05
Add ID extraction utilities for Notion URLs

riwaht:
This pull request introduces robust utilities for extracting Notion object IDs (pages, databases, blocks) from URLs or raw strings, and adds comprehensive tests for these helpers. The main focus is on reliably parsing Notion IDs from various URL formats and distinguishing between path and query parameter IDs, especially to avoid extracting view IDs instead of database IDs.

**ID Extraction Utilities:**

* Added new helper functions: `extract_notion_id`, `extract_database_id`, `extract_page_id`, and `extract_block_id` in `notion_client/helpers.py` for extracting and formatting Notion UUIDs from URLs or strings. These utilities prioritize extracting IDs from URL paths over query parameters to avoid common mistakes, such as extracting view IDs instead of database IDs. They also handle compact UUIDs, formatted UUIDs, and fragments for block IDs.
* Updated the `notion_client/__init__.py` exports to include the new ID extraction helpers, making them available as part of the public API.

**Testing Improvements:**

* Added a dedicated test class `TestIdExtractionUtilities` in `tests/test_helpers.py` with thorough test cases covering various Notion URL formats, compact and formatted UUIDs, domain validation, handling of whitespace and case, and block extraction from URL fragments. This ensures the new utilities work as intended across edge cases.
* Updated test imports to include the new helper functions.

**Internal Refactoring:**

* Refactored imports in `notion_client/helpers.py` to support new utilities.

### DIFF
--- a/notion_client/helpers.py
+++ b/notion_client/helpers.py
@@ -1,7 +1,16 @@
 """Utility functions for notion-sdk-py."""
 
 import re
-from typing import Any, AsyncGenerator, Awaitable, Callable, Dict, Generator, List, Optional
+from typing import (
+    Any,
+    AsyncGenerator,
+    Awaitable,
+    Callable,
+    Dict,
+    Generator,
+    List,
+    Optional,
+)
 from urllib.parse import urlparse
 from uuid import UUID
 
@@ -132,7 +141,6 @@ def _format_uuid(compact_uuid: str) -> str:
     if len(compact_uuid) != 32:
         raise ValueError("UUID must be exactly 32 characters")
 
-
     return (
         f"{compact_uuid[:8]}-{compact_uuid[8:12]}-{compact_uuid[12:16]}-"
         f"{compact_uuid[16:20]}-{compact_uuid[20:]}"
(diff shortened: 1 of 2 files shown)


notion_client/helpers.py
```diff
@@ -124,3 +134,169 @@ def is_equation_rich_text_item_response(rich_text: Dict[Any, Any]) -> bool:
...
+            # Remove "block-" prefix if present
+            if fragment.startswith("block-"):
+                fragment = fragment[6:]
+            if compact_uuid_pattern.match(fragment):
+                return _format_uuid(fragment.lower())
+
+        return None
+
+    except Exception:
+        return None
```
ramnes: Why is the implementation different than in notion-sdk-js here?

https://github.com/makenotion/notion-sdk-js/blob/6945289f774aa0150d9c382b1f88c752dbe0f925/src/helpers.ts#L249-L262

notion_client/helpers.py
```diff
@@ -124,3 +134,169 @@ def is_equation_rich_text_item_response(rich_text: Dict[Any, Any]) -> bool:
...
+
+    This is an alias for extract_notion_id for clarity when working with pages.
+
+    Args:
+        url_or_id: A Notion page URL or ID string
+
+    Returns:
+        The extracted UUID in standard format (with hyphens) or None if invalid
+    """
+    return extract_notion_id(url_or_id)
```
ramnes: I'm not sure if these are useful  but since they've been merged in notion-sdk-js, let's add them here as well. Let's just give the same argument names as in notion-sdk-js: `database_url` and `page_url`

notion_client/helpers.py
```diff
@@ -124,3 +134,169 @@ def is_equation_rich_text_item_response(rich_text: Dict[Any, Any]) -> bool:
...
+        The extracted UUID in standard format (with hyphens) or None if invalid
+
+    Example:
+        >>> # Database URL with view ID - extracts database ID, not view ID
+        >>> extract_notion_id('https://notion.so/workspace/DB-abc123def456789012345678901234ab?v=viewid123')
+        'abc123de-f456-7890-1234-5678901234ab'  # database ID
+
+        >>> # Already formatted UUID
+        >>> extract_notion_id('12345678-1234-1234-1234-123456789abc')
+        '12345678-1234-1234-1234-123456789abc'
```
ramnes: We use markdown for our docstrings, not Google-style docstrings. Can we remove the Args block, say "Returns" as a real sentence and add the example in a code block?

Let's do something similar for the other docstrings as well.

notion_client/helpers.py
```diff
@@ -124,3 +134,169 @@ def is_equation_rich_text_item_response(rich_text: Dict[Any, Any]) -> bool:
...
+                if uuid_pattern.match(fragment):
+                    return fragment.lower()
+
+            # If no valid fragment found, return None (block IDs should be in fragments)
+            return None
+        except Exception:
+            pass
+
+    # Fall back to general ID extraction for non-URL inputs
+    return extract_notion_id(url_or_id)
```
ramnes: Why is the implementation different than in notion-sdk-js? https://github.com/makenotion/notion-sdk-js/blob/6945289f774aa0150d9c382b1f88c752dbe0f925/src/helpers.ts#L304-L308

tests/test_helpers.py
```diff
@@ -236,3 +240,307 @@ def test_is_equation_rich_text_item_response(client, equation_block_id):
 def test_is_mention_rich_text_item_response(client, mention_block_id):
     response = client.blocks.retrieve(block_id=mention_block_id)
     assert is_mention_rich_text_item_response(response["paragraph"]["rich_text"][0])
+
+
+class TestIdExtractionUtilities:
```
ramnes: We don't use test classes so far, let's keep consistent and use pytest style

tests/test_helpers.py
```diff
@@ -236,3 +240,307 @@ def test_is_equation_rich_text_item_response(client, equation_block_id):
...
+    """Test the ID extraction utilities."""
+
+    def test_extract_notion_id_with_standard_urls(self):
+        """Test extracting ID from standard Notion URLs."""
```
ramnes: Let's remove all the docstrings that just repeat the test function name.

ramnes (changes requested):

notion_client/helpers.py
riwaht: They're mostly a developer quality of life addition, instead of using `extract_notion_id` for everything, it feels much simpler to have helpers for each ID. Fixing the argument names!

notion_client/helpers.py
riwaht: You're right, although it had more validation here, it was over-engineered. I will keep it to the same implementation as the JS SDK.

notion_client/helpers.py
riwaht: Same as above.

Issue #275, opened 2025-08-26
Implement the new data source and database endpoints for API version 2025-09-03


ramnes:
Notion is planning a major API update: https://github.com/makenotion/notion-sdk-js/pull/600

Let's replicate their work here.

FlorianWilhelm:
🙈 It seems there always a change with the Notion API. As far as I understand it only concerns database calls and is detailed [here](https://developers.notion.com/docs/upgrade-guide-2025-09-03). I won't be able to work on it but maybe it helps someone :-) I'll definitely use it in [Ultimate Notion](https://ultimate-notion.com) when it is implemented here.

sashanclrp:
In what status is this? 
Which branch should I clone to help you guys have this upgrade of the new Notion Database/Datasources API finished by the end of the week?

Please let me know!

FlorianWilhelm:
@sashanclrp as far as I know, no one has started to work on it.

jralduaveuthey:
@FlorianWilhelm You're absolutely right, and I apologize for the tone. Don't take my message too seriously - it was mostly Claude-generated with some context notes for my future reference.

I have huge respect for OSS maintainers and the work you do in your spare time. Thanks for the great work on this project!

PR #286, opened 2025-09-17
Added data source endpoints and refactored database endpoints

tsinglinrain:
1. Added `data source` endpoints and refactored `database` endpoints  
   1. For the `DataSourcesEndpoint` class, the official documentation is inconsistent in casing, using both `Data Source` and `data source`. I chose to follow the official documentation exactly as written.  
   2. For example, see the first line of [Retrieve a data source](https://developers.notion.com/reference/retrieve-a-data-source) and [Query a data source](https://developers.notion.com/reference/query-a-data-source).  
   3. In the `DataSourcesEndpoint` class, the order of `Body Params` follows the official JS SDK rather than the API reference documentation.  

2. Added support for `is_locked` in page `update`.  

3. Added `additional_data` in error responses.  

*(Note: This PR mainly focuses on data source and database updates, and does not fully cover all changes introduced in the latest TS version.)*  

4. Test coverage improvements:  
   1. Since the existing `tests\cassettes` used an old `notion-version`, I replaced them in bulk. However, the new version tests failed because `decode_compressed_response: True` caused VCR to attempt decompression, expecting a `body.string` format, while the new format uses `content`. Setting it to `False` resolved the issue.  
   2. My assumption: `vcrpy 5.1.0 + httpx 0.28.1` produces an unstable format mix, while `vcrpy 6.0.2 + httpx 0.28.1` consistently outputs the `body.string` format. Therefore, I upgraded `vcrpy` to version 6.0.2 (keeping `httpx 0.28.1`), and removed the previously written compatibility function. With this change, httpx automatically handles compression, so I set `decode_compressed_response: False`, ensuring cassette files retain their raw state — making debugging easier and reducing potential VCR-related bugs.  
   3. As a result, all `tests\cassettes` were regenerated, except for `tests\test_helpers.py`.  
   4. The expression `re.sub(RE_PAGE_ID, r1.uri, "")` was likely incorrect. In `re.sub(pattern, repl, string)`, the second parameter (`repl`) should be the replacement text, not the input string. I fixed this with a small change to improve robustness and error handling.  

related：https://github.com/ramnes/notion-sdk-py/issues/275


tsinglinrain:
`addition_data` and `request_id` are only supported in the new version (2025-09-03). It’s a bit tricky, so I plan to open another PR on top of this one after it gets merged.

FlorianWilhelm:
Any updates on this?

ramnes:
I'll review this week!

Issue #278, opened 2025-09-01
Replicate notion-sdk-js examples


ramnes:
The JS SDK has many nice examples: https://github.com/makenotion/notion-sdk-js/tree/4173ed8df18d666c7e313345d30b168523d895f4/examples

Let's replicate them here. :)

ramnes:
Examples moved there: https://github.com/makenotion/notion-cookbook

Related issue: https://github.com/makenotion/notion-cookbook/issues/3

rahulrao85:
I've submitted two PRs porting JS SDK examples to Python:

- **#361** — \parse-text-from-any-block-type\: extracts plain text from any block type (zero external deps, uses \iterate_paginated_api\)
- **#362** — \database-email-update\: polls database for status changes, sends email via SendGrid

Both follow existing example conventions (dotenv, interactive input fallback, README). Happy to iterate on feedback!

ramnes:
Reopening because I don't think we've ported all the examples here: https://github.com/makenotion/notion-cookbook/tree/main/examples

PR #292, opened 2025-10-26
Add generate random data example

Sumeet213:
Adds example for generating random test data in Notion databases.

  Python version of the JS SDK's generate-random-data example. Creates fake data using Faker library and demonstrates basic database operations.

  Related to #278


ramnes:
Great, thanks! There's some little things that are not exactly the same as in the JS SDK and I'd prefer that we stay as close as possible to the JS SDK. I'll merge as is but feel free to open another PR if you're interested in bringing this closer to the original implementation. :)

PR #314, opened 2026-02-13
Enhance request handling in `client.py` following notion-sdk-js

tsinglinrain:
## Summary

- Imported `is_http_response_error`, `is_notion_client_error`, and `validate_request_path` in `client.py`, aligning with `notion-sdk-js`  
- Added `_log_request_success` and `_log_request_error` functions modeled after `_parse_response` to reduce code duplication between `Client` and `AsyncClient`  
- Introduced `TypeGuard` in `errors.py`; did not fully replicate all type guards from `notion-sdk-js`, only added what was necessary to satisfy `mypy` type checking  
- Did not adopt `RequestTimeoutError.reject_after_timeout()` from `errors.py`; continued using `httpx.TimeoutException` for a more Pythonic implementation

### DIFF
--- a/notion_client/client.py
+++ b/notion_client/client.py
@@ -251,17 +251,13 @@ class Client(BaseClient):
     ) -> Any:
         """Send an HTTP request."""
         request = self._build_request(method, path, query, body, form_data, auth)
-        return self._execute(request, method, path, query, body, form_data, auth)
+        return self._execute(request, method, path)
 
     def _execute(
         self,
         request: Request,
         method: str,
         path: str,
-        query: Optional[Dict[Any, Any]],
-        body: Optional[Dict[Any, Any]],
-        form_data: Optional[Dict[Any, Any]],
-        auth: Optional[Union[str, Dict[str, str]]],
     ) -> Any:
         """Executes the request with retry logic."""
         try:
@@ -328,17 +324,13 @@ class AsyncClient(BaseClient):
     ) -> Any:
         """Send an HTTP request asynchronously."""
         request = self._build_request(method, path, query, body, form_data, auth)
-        return await self._execute(request, method, path, query, body, form_data, auth)
+        return await self._execute(request, method, path)
 
     async def _execute(
         self,
         request: Request,
         method: str,
         path: str,
-        query: Optional[Dict[Any, Any]],
-        body: Optional[Dict[Any, Any]],
-        form_data: Optional[Dict[Any, Any]],
-        auth: Optional[Union[str, Dict[str, str]]],
     ) -> Any:
         """Executes the request with retry logic."""
         try:


notion_client/client.py
```diff
@@ -160,10 +164,33 @@ def _parse_response(self, response: Response) -> Any:
...
+        request_id = response_body.get("request_id")
+        if request_id:
+            self.logger.info(
+                f"request success: method={method}, path={path}, "
+                f"request_id={request_id}"
+            )
+        else:
+            self.logger.info(f"request success: method={method}, path={path}")
+
+    def _log_request_error(self, error: Exception) -> None:
```
ramnes: The JS SDK doesn't throw in `logRequestError`, and I don't think we should either? I think we should re-raise at the caller level.

notion_client/client.py
```diff
@@ -224,10 +251,15 @@ def request(
...
-        except httpx.TimeoutException:
-            raise RequestTimeoutError()
-        return self._parse_response(response)
+            try:
+                response = self.client.send(request)
+            except httpx.TimeoutException:
+                raise RequestTimeoutError()
+            response_body = self._parse_response(response)
+            self._log_request_success(method, path, response_body)
+            return response_body
```
ramnes: I think we should move this outside of the try block

notion_client/client.py
ramnes: Let's do the error type check in the caller as well

notion_client/errors.py
```diff
@@ -7,6 +7,7 @@
 import json
 from enum import Enum
 from typing import Any, Dict, Optional, Union, Set
+from typing_extensions import TypeGuard
```
ramnes: Let's use a conditional import so that >=3.10 directly imports from the standard library

setup.py
```diff
@@ -19,6 +19,7 @@ def get_description():
     python_requires=">=3.8, <4",
     install_requires=[
         "httpx >= 0.23.0",
+        "typing_extensions >= 4.0.0",
```
ramnes: Let's only depend on the backport if python < 3.10

tests/test_client.py
```diff
@@ -184,3 +184,38 @@ async def test_async_client_request_timeout():
...
+        mock_info.assert_called_once_with("request success: method=GET, path=/users")
+
+
+def test_log_request_error_with_non_notion_error(client):
+    """Test _handle_request_error with non-notion error."""
+
+    generic_error = ValueError("Some generic error")
+
+    with pytest.raises(ValueError, match="Some generic error"):
+        client._log_request_error(generic_error)
```
ramnes: Can we avoid testing private methods and instead test the whole workflow end to end? Feel free to discard if it's not convenient.

notion_client/client.py
```diff
@@ -160,10 +164,33 @@ def _parse_response(self, response: Response) -> Any:
...
+            )
+        else:
+            self.logger.info(f"request success: method={method}, path={path}")
+
+    def _log_request_error(self, error: Exception) -> None:
+        """Logs a request error with appropriate detail level."""
+        if not is_notion_client_error(error):
+            raise error
+
+        # Log the error if it's one of our known error types
```
ramnes: I appreciate the effort to make the code easy to understand with comments but I think that it's pretty obvious here, so let's remove that kind of comment that hinders the code readability without providing much value.

notion_client/errors.py
tsinglinrain:
```python
if sys.version_info >= (3, 10):
    from typing import TypeGuard
else:
    from typing_extensions import TypeGuard
```

setup.py
tsinglinrain:
```python
    install_requires=[
        "httpx >= 0.23.0",
        "typing_extensions >= 4.0.0; python_version < '3.10'",
    ],
```

tests/test_client.py
tsinglinrain: The new tests exercise the full request workflow by mocking client.client.send and asserting on observable outcomes.

Issue #313, opened 2026-02-04
Add automatic retry support with exponential backoff


ramnes:
notion-sdk-js now implements retries: https://github.com/makenotion/notion-sdk-js/pull/664

Related issue and discussion: #262

Issue #283, opened 2025-09-09
Export all things in the main namespace


ramnes:
The JS SDK exports many things in the top level namespace. Let's do the same here once #282 is merged.

https://github.com/makenotion/notion-sdk-js/blob/main/src/index.ts

PR #319, opened 2026-03-07
Add automatic retry support with exponential backoff

tsinglinrain:
## Summary

Implements a comprehensive automatic retry mechanism in `client.py`, mirroring the JavaScript SDK. Supports exponential backoff and `Retry-After` response header parsing.

Exports additional public namespaces from `__init__.py` for easier access to related types by users.

Updates `README.md` with retry-related usage instructions and configuration documentation, following the JS version as reference.

Adds extensive test cases to `test_client.py`, introducing the `_mock_http_response()` helper to reduce boilerplate and convenience factories like `success_response()` to improve test readability.

Adds a `RetryOptions` dataclass to centralize retry-related configuration.

Related issues: https://github.com/ramnes/notion-sdk-py/issues/283, https://github.com/ramnes/notion-sdk-py/issues/313

## Differences from the JS Version

The JS SDK uses `Date.parse(retryAfterValue)` to parse HTTP date strings, which is natively supported without any extra imports.

Python's standard library lacks a dedicated HTTP date parser, but `email.utils.parsedate_to_datetime` serves the purpose perfectly — HTTP-date and email date headers both follow the same RFC 2822 specification. A rather elegant reuse of existing infrastructure.

### DIFF
--- a/notion_client/client.py
+++ b/notion_client/client.py
@@ -155,7 +155,6 @@ class BaseClient:
         auth: Optional[Union[str, Dict[str, str]]] = None,
     ) -> Request:
         headers = httpx.Headers()
-        validate_request_path(path)
         if auth:
             if isinstance(auth, dict):
                 client_id = auth.get("client_id", "")
@@ -165,8 +164,6 @@ class BaseClient:
                 headers["Authorization"] = f"Basic {encoded_credentials}"
             else:
                 headers["Authorization"] = f"Bearer {auth}"
-        self.logger.info(f"{method} {self.client.base_url}{path}")
-        self.logger.debug(f"=> {query} -- {body} -- {form_data}")
 
         if not form_data:
             return self.client.build_request(
@@ -361,6 +358,8 @@ class Client(BaseClient):
         auth: Optional[Union[str, Dict[str, str]]] = None,
     ) -> Any:
         """Send an HTTP request."""
+        validate_request_path(path)
+        self.logger.info(f"{method} {self.client.base_url}{path}")
         return self._execute_with_retry(method, path, query, body, form_data, auth)
 
     def _execute_with_retry(
@@ -448,6 +447,8 @@ class AsyncClient(BaseClient):
         auth: Optional[Union[str, Dict[str, str]]] = None,
     ) -> Any:
         """Send an HTTP request asynchronously."""
+        validate_request_path(path)
+        self.logger.info(f"{method} {self.client.base_url}{path}")
         return await self._execute_with_retry(
             method, path, query, body, form_data, auth
         )


notion_client/client.py
```diff
@@ -324,23 +456,44 @@ async def request(
...
-    async def _execute(
+    async def _execute_with_retry(
         self,
         request: Request,
         method: str,
         path: str,
+        query: Optional[Dict[Any, Any]],
+        body: Optional[Dict[Any, Any]],
+        form_data: Optional[Dict[Any, Any]],
+        auth: Optional[Union[str, Dict[str, str]]],
```
ramnes, line 463: I think the signature could be improved here? It seems to me that we should either reuse request in the loop, or on the opposite always build the request. Right now we're both using `request` from the argument and then rebuilding a new one in the loop, it's a bit weird. Is it really necessary?

notion_client/client.py
```diff
@@ -183,16 +223,87 @@ def _log_request_success(self, method: str, path: str, response_body: Any) -> No
...
+                APIErrorCode.ServiceUnavailable,
+            )
+
+        return False
+
+    def _calculate_retry_delay(self, error: Exception, attempt: int) -> float:
+        """Calculates the delay before the next retry attempt.
+
+        Uses retry-after header if present, otherwise exponential back-off with jitter.
+        Returns delay in seconds.
```
ramnes, line 261: The variable is called `delay_ms` but it's seconds? Something is weird here.

notion_client/client.py
```diff
@@ -251,23 +362,44 @@ def request(
...
+        while True:
+            try:
+                return self._execute_single_request(request, method, path)
+            except Exception as error:
+                if not is_notion_client_error(error):
+                    raise error
+
+                self._log_request_error(error, attempt)
+
+                if attempt < self._max_retries and self._can_retry(error, method):
```
ramnes: Can we please do the opposite here and re-raise / early-return if we don't satisfy these? So that we can remove the `continue` logic that's harder to read.

notion_client/client.py
tsinglinrain: Good catch. The variable was named `delay_ms`, but it had already been divided by 1000, so it was effectively in seconds. I renamed it to `delay` to reflect the correct unit. The division by 1000 is kept so it can be passed directly to `time.sleep(delay)`.

notion_client/client.py
tsinglinrain: Good point! Inverted the condition to raise early when we shouldn't retry,  so the retry logic falls through naturally without needing continue.
```python
if attempt >= self._max_retries or not self._can_retry(error, method):
    raise error

delay = self._calculate_retry_delay(error, attempt)
self.logger.info(
    f"retrying request: method={method}, path={path}, attempt={attempt + 1}, delay_ms={delay * 1000:.0f}"
)
time.sleep(delay)
attempt += 1
```

notion_client/client.py
tsinglinrain: Good catch. The original code built the request outside the loop and passed it in alongside the raw parameters, then redundantly rebuilt it on each retry — leaving both the pre-built `request` and the rebuild parameters in the signature at the same time.

Fixed by removing the `request` parameter from `_execute_with_retry` entirely and always building it at the top of the loop, which aligns with how the JS version handles this.
```python
def request(...) -> Any:
    """Send an HTTP request."""
    # ...
    return self._execute_with_retry(method, path, query, body, form_data, auth)
```

```python
def _execute_with_retry(...) -> Any:
    """Executes the request with retry logic."""
    attempt = 0
    while True:
        request = self._build_request(method, path, query, body, form_data, auth)
        try:
            return self._execute_single_request(request, method, path)
        except Exception as error:
            # ...
```

tsinglinrain: Before this change, the `request()` method called `_build_request()`, which indirectly invoked `validate_request_path() `to validate the request path.

After introducing `_execute_with_retr`y, `request()` no longer calls `_build_request()` directly. As a result, `validate_request_path()` would be executed on every retry attempt instead of only once at the request entry point.

To preserve the original behavior, I moved `validate_request_path()` to the beginning of the `request()` method so that the validation happens only once per request.

Regarding logging: `_log_request_success` and `_log_request_error` already cover the full lifecycle of a request. The debug log
`self.logger.debug(f"=> {query} -- {body} -- {form_data}")`
did not provide much additional value, so I removed it.

Since `_build_request()` is now only responsible for constructing the request object, I also moved
`self.logger.info(f"{method} {self.client.base_url}{path}")`
to the `request()` method. This makes the control flow clearer and keeps request-level logging at the entry point.

The updated structure looks like this:
```python
def request(...):
    validate_request_path(path)  # moved here
    self.logger.info(f"{method} {self.client.base_url}{path}")  # moved here
    return self._execute_with_retry(method, path, query, body, form_data, auth)
```

notion_client/client.py
ramnes: should we keep delay in milliseconds internally to match JS SDK, convert only at `sleep()` call?

tsinglinrain: I agree with you, that's more cleaner. Keeping `delay_ms` in milliseconds internally is consistent with both the JS SDK and the existing parameters in `RetryOptions` class.

I’d be happy to address it in a follow-up PR: keep delay_ms in milliseconds internally and convert only at the `sleep()` call site.

PR #349, opened 2026-06-30
Handle trailing slashes in get_id URLs

CodingFeng101:
## What changed

`get_id()` now strips trailing slashes from the parsed URL path before taking the final 32-character Notion ID segment.

## Why

Notion page URLs are commonly copied with a trailing `/`. The previous implementation included that slash in the final 32 characters, which caused `UUID(...)` validation to fail even though the URL otherwise contained a valid page ID.

## Validation

- `python -m pytest tests\test_helpers.py::test_get_id -q --no-cov`
- `python -m compileall -q notion_client tests\test_helpers.py`
- `git diff --check`

I also ran `python -m flake8 notion_client\helpers.py tests\test_helpers.py`; it reports pre-existing E501 long lines in these files that are unrelated to this change.

### DIFF
--- a/notion_client/helpers.py
+++ b/notion_client/helpers.py
@@ -38,7 +38,7 @@ def get_id(url: str) -> str:
     parsed = urlparse(url)
     if parsed.netloc not in ("notion.so", "www.notion.so"):
         raise ValueError("Not a valid Notion URL.")
-    path = parsed.path
+    path = parsed.path.rstrip("/")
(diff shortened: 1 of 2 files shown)


ramnes (approved):
LGTM. We probably should deprecate that `get_id` in favor of `extract_notion_id` but that's another issue. :)

ramnes:
CI errors seem unrelated, merging.

PR #352, opened 2026-07-02
Add httpx2 compatibility

### LINKED ISSUE #344: Consider migrating from `httpx` to `httpx2` (Pydantic's maintained fork)
[**`httpx`**](https://github.com/encode/httpx), this SDK's HTTP backend (`setup.py`: `httpx >= 0.23.0`), appears to have stalled: the last **stable** release was 0.28.1 (Dec 2024), with only `1.0.dev*` pre-releases since, and community channels (issues/discussions) being closed.

In May 2026, the Pydantic team published [**`httpx2`**](https://github.com/pydantic/httpx2), a fork that explicitly takes over stewardship:

> With HTTPX itself seeing limited activity recently, Pydantic is picking up stewardship under the HTTPX2 name so that users have a reliably maintained path forward — including timely security updates for a library that sits in the critical path of so many production systems...

Only one has been found so far, and it seems that no more projects have been migrated:

- Starlette merged its migration (encode/starlette#3291)
- fastapi add httpx2 test dependency to avoid deprecation warning (fastapi/fastapi#15603)

## Impact on this project

The good news is that the surface area here is small. This repo uses `httpx` in a fairly standard, narrow way (`httpx.Client` / `httpx.AsyncClient`, `httpx.Headers`, request calls, exception types), and `httpx2` is advertised as a drop-in successor, so the bulk of a migration would be a mechanical `httpx` → `httpx2` substitution.

**1. Python version floor.** `httpx2` requires [Python 3.10+](https://github.com/pydantic/httpx2/blob/main/uv.lock#L3). This SDK currently advertises support down to 3.8 (`python_requires=">=3.8, <4"`, with 3.8/3.9 classifiers). Adopting `httpx2` would force dropping 3.8 and 3.9.

**2. The public "bring your own client" API.** The SDK lets users pass their own `httpx.Client` / `httpx.AsyncClient` into the constructor. This is the one place where `httpx` types leak into the public interface. If we switch the internal dependency to `httpx2`, a user-supplied `httpx.Client` is now a *different type* from the SDK's internal `httpx2.Client`.

**3. Test suite.** `httpx2`'s underlying transport is `httpcore2` rather than `httpcore`. The test suite records HTTP interactions (VCR cassettes); a transport change can affect request matching, so cassettes may need re-recording or matcher adjustments.

**4. Downstream churn.** Whatever we choose propagates to every consumer's dependency tree. A hard switch is simplest for maintainers but pushes a `httpx`/`httpx2` decision (and the 3.10+ floor) onto all downstream users at once.

tsinglinrain:
## Summary

Adds support for [httpx2](https://github.com/pydantic/httpx2) (Pydantic's maintained continuation of httpx) as an **optional, opt-in** HTTP backend, while keeping `httpx` as the default dependency. `BaseClient` / `Client` / `AsyncClient` are now library-agnostic: they work with either backend, and no user is forced to have both installed. 

Resolves #344.

## What changed

- **`notion_client/compat.py` (new):** the single module that names `httpx`/`httpx2`. It probes for both at import time (httpx2 preferred when present), re-exports the active backend's `Client`/`AsyncClient`/`Request`/`Response`/`Headers`, and builds exception tuples (`HTTP_STATUS_ERRORS`, `TIMEOUT_EXCEPTIONS`) spanning **every**  installed backend so a user-supplied client from either library is handled.  `NOTION_HTTP_BACKEND` can force a backend (mainly for tests/CI).
- **`client.py` / `errors.py`:** import from `compat` instead of a concrete library; the client is configured with neutral values (str/float/dict) rather than backend-specific `URL`/`Timeout`/`Headers` objects.
- **Packaging (`setup.py`):** base install keeps `httpx`; adds `extras_require={"httpx2": ["httpx2 >= 2.0.0"]}`. Python 3.8+ support retained.
- **Testing:** tox matrix covers both backends (`py3{8,9,10,11,12,13,14}-httpx` + `py3{10,11,12,13,14}-httpx2`); shared plugins factored into `requirements/test-plugins.txt`; [vcrpy 8.2.0](https://github.com/kevin1024/vcrpy/releases/tag/v8.2.0) (native httpx2 support, Python ≥ 3.10) with a fallback to 6.0.2 so the httpx path stays tested on 3.8/3.9. Cassettes are HTTP-level and replay unchanged on both backends.


requirements/tests.txt
```diff
@@ -1,8 +1,3 @@
...
-pytest
-pytest-asyncio
-pytest-cov
-pytest-mock
-pytest-timeout
-pytest-vcr
-vcrpy==6.0.2
+-e .[httpx2]
+-r test-plugins.txt
+vcrpy==8.2.0
```
ramnes: I don't get it, why do we not just add 

```
httpx2
vcrpy==8.2.0; python_version >= "3.10"
vcrpy==6.0.2; python_version < "3.10"
```

to the original file and use the same file for all tests?

I don't like the three files too much

README.md
```diff
@@ -453,6 +453,11 @@ This package supports the following minimum versions:
 - Runtime: Python >= 3.8
 - httpx >= 0.23.0
 
+The client is HTTP-backend agnostic: it works with either `httpx` (the default)
```
ramnes, line 456: I wouldn't say it's entirely agnostic, we don't support requests for example

ramnes (commented):
LGTM overall with a few minor nits (see comments), but more importantly, can we make sure that users that want to use httpx2 don't pull httpx?

ramnes:
(also please update on main to have CI fixed)

tsinglinrain:
> can we make sure that users that want to use httpx2 don't pull httpx?

@ramnes Thanks for the review!

That was my initial thought, but with plain setuptools extras it isn't directly possible: an extra can only *add* dependencies, so as long as `httpx` stays in `install_requires`, installing `[httpx2]` will always drag `httpx` in too.

Making `[httpx2]` not pull `httpx` and this needs a small design decision. One approach is to move `httpx` out of `install_requires` and make both backends extras:
- `pip install notion-client[httpx]` → httpx only (the documented default)
- `pip install notion-client[httpx2]` → httpx2 only, no httpx

However, the impact of this is that a bare `pip install notion-client` would no longer install a backend on its own. `compat.py` already raises a clear `ImportError` telling the user to install one of the two, so it fails loudly with guidance rather than at first request. Are you OK with that trade-off? Personally, I think completely separating `httpx2` and `httpx` is the more elegant and appropriate approach (which I agree with you), and the resulting impact is acceptable.

ramnes:
Sorry for the delay, I needed a bit more time to think this through.

I don’t want to keep httpx as a required dependency while pretending httpx2 is an alternative backend. That feels half-baked to me.

I think we have the following options here:

* Do nothing for now and wait (or push!) for [PEP 771](https://peps.python.org/pep-0771/), which would let us keep httpx as the default without forcing it on httpx2 users.
* Make the SDK truly transport-free, in a new major version.
* Release a separate `notion-sdk-py-httpx2` package.

If we choose the second option, we should first cut a last release of the current major, then merge the breaking transport changes. We should also design this as a real abstraction and support more than just httpx2, for example requests and aiohttp. But even then, I'm not too sure I want to go this route: httpx is still the most used HTTP library after urllib3 and requests...

The third option could give httpx2 users a clean dependency set without changing the existing package, but it's a bit unusual / ugly and would mean maintaining another distribution, i.e. a little more work on me.

What do you think?

Right now I'd rather not merge this. This is really about the direction and packaging trade-offs, not the quality of the code itself by the way, thanks again for putting the work into it! :)

tsinglinrain:
@ramnes Thanks for taking the time to think this through, and for the thorough review — I really appreciate it.

I agree with your core concern: keeping `httpx` as a required dependency while presenting `httpx2` as an optional backend is half-baked, and I'd rather not ship it that way either. Of the three options, option 1 (waiting on / nudging PEP 771) looks cleanest to me too — it's the only one that keeps `httpx` as the zero-config default while giving httpx2-only users a clean dependency set, and without a second package to maintain.

I'm happy to defer to you on how to handle this PR —  keep it open as a draft for reference, or close it , whatever you prefer works for me. I'd be glad to keep helping.

ramnes:
Let's keep it open for now and see if people react to that discussion, or if PEP 771 gets some significant movement.

PR #375, opened 2026-08-09
Match the async paginator's termination to the sync one

ckarnell:
`async_iterate_paginated_api` terminates on `(not response["has_more"]) | (next_cursor is None)`. The sync twin twenty lines up uses `not response.get("has_more") or not next_cursor`, and so do both data source template iterators further down, so three of the four agree and this one is the odd one out.

Two behaviours differ. A response without `has_more` stops the sync iterator and raises `KeyError` on the async one. An empty string cursor stops sync and loops forever on async, because `"" is None` is False.

This looks like drift, not a decision. Both forms arrived in the same commit with sync directly above async, and a later commit titled "Fix iteration helpers" edited the lines immediately above both terminators and left this one alone.

Honest about the severity. I have no Notion account, so I have not shown that the live API ever omits `has_more` or hands back an empty cursor. What I can show is that the two halves of the same helper disagree about it, and that the async half is the one that fails badly where the other fails safe.

`tests/test_async_paginator_termination.py` runs each case through both iterators. Both tests fail on `main`, one with `KeyError: 'has_more'` and the other by never terminating, and both pass with the change. The full suite is 194 passed. Coverage stays where CONTRIBUTING asks.


ramnes:
Oof. :) 

Can you just put the tests in `tests/test_helpers.py` and drop the `@pytest.mark.asyncio` (we use `asyncio_mode = auto` and shouldn't need it)?

Thanks!

ramnes:
Also, even after merging this, all four Python paginators check `has_more`, which the JS SDK does not; feel free to do the larger update if you feel like it!

ckarnell:
Done, both. The tests are in `tests/test_helpers.py` with the rest of the helper tests and the markers are gone, since `setup.cfg` already sets `asyncio_mode = auto`. The local fake is renamed `_paginator_pages` now it shares a module.

Suite is 194 passed. I also reverted the source fix to check the move had not quietly turned them into tests that pass no matter what, and both still fail without it.

On the `has_more` check across all four paginators, yes, I would like to. Same for the path-match one on #376. I will do them one at a time instead of sending you three branches at once. Say which you would rather see first and I will start there.

ramnes:
Thanks, merged! 

Let's do a single PR for the three remaining `has_more` checks please, so that it reduces my bandwidth a little bit :)

ckarnell:
Done in #377. I did four instead of three and said why in the description.

PR #376, opened 2026-08-09
Accept notion.com in the URL helpers

ckarnell:
`get_id` accepts only `notion.so` and `www.notion.so`, and `extract_notion_id` only matches `notion.so` or `notion.site`. Notion serves object URLs from `app.notion.com`, so `get_id(page["url"])` raises `ValueError` on the URL the API just handed you, and `extract_notion_id` returns `None` for it without saying why.

The evidence is in this repo. Every object URL in the recorded cassettes is on `app.notion.com`:

```
$ grep -ohE '"(url|public_url)":"https://[^/"]+' tests/cassettes/*.yaml | sed 's|.*https://||' | sort | uniq -c
  98 app.notion.com
```

Those same files carried `www.notion.so` before the cassettes were re-recorded against the live API in July. The allowlist has not been touched since "Fix get_url (#80)", so this reads as the host moving underneath it, not a deliberate restriction.

Both helpers now share `_is_notion_host`, which accepts `notion.com`, `notion.so` and `notion.site` and any subdomain of them.

The part worth reviewing is that this does not become a suffix match. `notion.so.evil.com`, `notion.com.evil.com`, `notsonotion.com` and `evil.com` are all rejected before and after, and those cases are in the tests so the guarantee is pinned, not argued. `extract_notion_id` still returns `None` instead of raising on unparseable input, which is the contract `test_extract_block_id_url_parsing_exception` already covers.

One test asserts the cassettes really are all on `app.notion.com`, so if that ever changes this fails loudly instead of passing for the wrong reason.

Three of the ten new tests fail on `main` and all ten pass here. Full suite is 202. Coverage on `helpers.py` stays at 100%, and `get_url()` is untouched, so nothing changes about the host we emit.

I do not have a Notion account, so this rests on your recorded responses, not on live calls I made.

### DIFF
--- a/notion_client/helpers.py
+++ b/notion_client/helpers.py
@@ -33,10 +33,22 @@ def get_url(object_id: str) -> str:
     return f"https://notion.so/{UUID(object_id).hex}"
 
 
+#: Registrable domains Notion serves object URLs from. `notion.com` is the one the API
+#: returns today (`page["url"]` is `https://app.notion.com/...`); `notion.so` is the legacy
+#: host and still resolves; `notion.site` serves published pages.
+_NOTION_DOMAINS = ("notion.com", "notion.so", "notion.site")
+
+
+def _is_notion_host(netloc: str) -> bool:
+    """Return `True` for a Notion domain or any of its subdomains (app., www., ...)."""
+    host = netloc.lower().partition(":")[0]
+    return any(host == d or host.endswith(f".{d}") for d in _NOTION_DOMAINS)
+
+
 def get_id(url: str) -> str:
     """Return the id of the object behind the given URL."""
     parsed = urlparse(url)
-    if parsed.netloc not in ("notion.so", "www.notion.so"):
+    if not _is_notion_host(parsed.netloc):
         raise ValueError("Not a valid Notion URL.")
     path = parsed.path.rstrip("/")
     if len(path) < 32:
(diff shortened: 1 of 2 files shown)


ramnes:
Good catch, thanks!

The JS SDK has no domain check at all in `extractNotionId`, can we remove it entirely and stay closer to the original implementation?

It looks like we're also not doing the path-match step right, if you want to handle this (or we can open a separate issue):

```
https://notion.so/ws/My-Page-<A>?p=<B>
  js: <A>   (pathMatch runs before queryMatch)
  py: <B>   (goes straight to the query fallback)
```

Last but not least, Code Quality CI is failing, install pre-commit if you want to make sure it passes next time :)

ckarnell:
Yes, remove it. That deletes the reason for this PR instead of patching it, which I prefer.

The only thing the host check buys is returning `None` for a non-Notion URL that happens to contain 32 hex characters. Matching `extractNotionId` is worth more than that, and it removes a list someone has to remember to update the next time Notion adds a domain, which is how this came up.

On the path-match step, you are right and it is a bit worse than the example shows. Three things I found while reproducing it.

There is no path-match step in the function at all. The order is bare UUID, compact UUID, host check, query match, then any 32-hex.

The docstring claims the opposite: "Prioritizes path IDs over query parameters to avoid extracting view IDs instead of database IDs."

And the docstring's own example only passes by accident. It uses `?v=viewid123`, and `v` is not in the query pattern, which is `p|page_id|database_id`. So the query branch misses, it falls through to the any-32-hex scan, and that happens to hit the path id first because it appears earlier in the string. Swap `?v=` for `?p=` and the same call returns the query id:

```
DB-<A>?v=viewid123      -> <A>   path id, the documented behaviour, by luck
DB-<A>?p=<B>            -> <B>   query id
DB-<A>?database_id=<B>  -> <B>   query id
```

So it is documented behaviour the code never implemented, and the example that would have caught it is the one input where the missing branch does not matter.

Happy to take it. Separate PR though, since it changes what the function returns and this one only widens what it accepts.

On Code Quality, that is fixed and pushed already. The pinned ruff-format wanted the assert in the new test reshaped. The run has not gone green because the workflows are sitting on `action_required` and need approving.

PR #377, opened 2026-08-10
Terminate the paginators on the cursor alone, as the JS SDK does

ckarnell:
The single PR you asked for, following #375.

`iteratePaginatedAPI` in notion-sdk-js loops on `while (nextCursor)` and never reads `has_more`. All four Python paginators required it, so a response carrying a cursor with `has_more` false stopped here and kept going there.

One line each. `iterate_paginated_api`, `async_iterate_paginated_api`, `iterate_data_source_templates` and `async_iterate_data_source_templates`.

That is four, not the three you counted, and the extra one is `async_iterate_paginated_api` from #375. Dropping the check from the other three while leaving it there would put the async paginator back out of step with the sync one, which is the thing #375 closed. It seemed worse than the wider diff, but it is one line and you should revert it if you would rather keep them separate.

Both tests from #375 still pass untouched. They turn on an empty-string cursor, which is falsy with or without the `has_more` clause, so the fix you just merged is unaffected.

Added one test for the behaviour this actually changes: a page with `has_more` false and a cursor set is now followed, in both the sync and async paginator. It fails on `main` and passes here. Suite is 195, up from 194.

I read the JS side myself instead of taking it from your comment, since this deletes a guard. `src/helpers.ts`, the `do/while (nextCursor)` loop.

### DIFF
--- a/notion_client/helpers.py
+++ b/notion_client/helpers.py
@@ -57,7 +57,7 @@ def iterate_paginated_api(
             yield result
 
         next_cursor = response.get("next_cursor")
-        if not response.get("has_more") or not next_cursor:
+        if not next_cursor:
             return
 
 
@@ -78,7 +78,7 @@ async def async_iterate_paginated_api(
             yield result
 
         next_cursor = response.get("next_cursor")
-        if not response.get("has_more") or not next_cursor:
+        if not next_cursor:
             return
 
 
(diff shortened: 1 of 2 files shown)


ramnes (approved):

ramnes:
Sorry for the delay here, and thanks for your contribution!

PR #390, opened 2026-09-06
Prevent secrets in pull request tests

andife:
Remove Notion and Codecov secrets from pull-request execution. Live integration tests and coverage uploads continue only on trusted pushes to main, preventing PR code from accessing repository credentials.


ramnes:
Why? It means we won't be able to see coverage updates on PRs anymore, right? Also I think the Notion token isn't needed anymore since we use cassettes, so we should probably remove it entirely.

PR #391, opened 2026-09-06
Harden workflow permissions and action pinning

andife:
Declare least-privilege GitHub Actions permissions, grant write access only to workflows that require it, and pin the remaining mutable checkout action reference.


ramnes (approved):
LGTM, thanks!

PR #393, opened 2026-09-06
Fix `is_full_data_source` accepting partial responses

tsinglinrain:
Related pr: makenotion/notion-sdk-js#764, related issue: https://github.com/ramnes/notion-sdk-py/issues/380

`is_full_data_source` only checked `response["object"]`, so a partial data source response passed the guard:

```python
partial = {"object": "data_source", "id": "...", "properties": {}}
is_full_data_source(partial)  # was True, now False
```

It now also requires `"title" in response`, matching the structural check the other guards already do (`is_full_page` checks `url`, `is_full_block` checks`type`). `is_full_page_or_data_source` picks up the fix through `is_full_data_source`.

### Why only one guard here

The JS PR fixes both `isFullDataSource` and `isFullDatabase`. Only the former needs it in this SDK.

In the JS SDK, `isFullDatabase` did have the `"title"` check up to and including v4.0.2. It was dropped in makenotion/notion-sdk-js#600, the refactor that split databases into databases and data sources — that commit both removed the check from `isFullDatabase` and introduced `isFullDataSource` without one: [changes](https://github.com/makenotion/notion-sdk-js/pull/600/changes#diff-ca5696b367d474e785317cb7a0d9853fef5729387ab8d0fab4c46268c03aae99R135-R138).

I ported the data source helpers here in #299, about two months later, and mirrored the JS implementation as it_source`
inherited the missing check. `is_full_database` was left untouched in that PR and kept the `"title"` check it has carried since #148 (2022), so it never followed the upstream regression and this brings the Python side in line. I'm sorry, I forgot to make the changes.

### Why the existing tests didn't catch it

All the `is_full_*` tests are VCR-backed and were recorded with an integration holding full read capabilities, so every cassette contains complete objects.

I only found this passage in the official documentation.
[working-with-comments](https://developers.notion.com/guides/data-apis/working-with-comments)

> The exception to what will be returned occurs if your connection has “write comment” capabilities but not “read comment” capabilities. In this situation, the response will be a partial object consisting of only the id and object fields. This is because the connection can create new comments but can’t retrieve comments, even if the retrieval is just the response for the newly created one.

### Verified against a live integration

Therefore, following the hints provided in the aforementioned document, I conducted actual tests. Reproduced with an integration that has no "Read content" capability. The API returns HTTP 200 with a partial object rather than erroring:
    {"object": "data_source", "id": "...", "properties": {}}

    GET /v1/databases/{id}
    {"object": "database", "id": "..."}

These match the asterisked fields on the data source and database references exactly. Before the fix, `is_full_data_source()` returned `True` for that first response and `response["title"]` raised `KeyError`.

### DIFF
--- a/notion_client/helpers.py
+++ b/notion_client/helpers.py
@@ -100,8 +100,8 @@ def is_full_page(response: Dict[Any, Any]) -> bool:
 
 
 def is_full_data_source(response: Dict[Any, Any]) -> bool:
-    """* Return `true` if `response` is a full data source."""
-    return response.get("object") == "data_source"
+    """Return `True` if response is a full data source."""
+    return response.get("object") == "data_source" and "title" in response
 
 
 def is_full_database(response: Dict[Any, Any]) -> bool:
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -187,6 +187,37 @@ def test_is_full_page_or_data_source(client, data_source_id, page_id):
     assert is_full_page_or_data_source(response)
 
 
+def test_is_full_data_source_with_partial_response():
+    partial = {"object": "data_source", "id": "data-source-id", "properties": {}}
+    assert not is_full_data_source(partial)
+
+
+def test_is_full_database_with_partial_response():
+    partial = {"object": "database", "id": "database-id"}
+    assert not is_full_database(partial)
+
+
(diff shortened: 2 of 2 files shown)


ramnes (approved):
LGTM

PR #394, opened 2026-09-09
Add `service_overload` (HTTP 529) retry support

tsinglinrain:
## Description

Adds first-class handling for `service_overload` API responses so HTTP 529 errors are parsed as `APIResponseError`s and retried using the same contract as `rate_limited` (429). 529 responses are retried for all HTTP methods and respect the `Retry-After` header.

Before this change, a 529 body fell through to `UnknownHTTPResponseError` because `service_overload` was not a known `APIErrorCode`, so it was never retried.

Aligns with makenotion/notion-sdk-js#727.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ These options are all keys in the single constructor parameter.
 | `timeout_ms` | `DEFAULT_TIMEOUT_MS`        | `int`             | Number of milliseconds to wait before emitting a `RequestTimeoutError`                                                                    |
 | `base_url`   | `DEFAULT_BASE_URL`          | `string`          | The root URL for sending API requests. This can be changed to test with a mock server.                                                    |
 | `logger`     | Log to console              | `logging.Logger`  | A custom logger.                                                                                                                          |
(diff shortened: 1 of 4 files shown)


tsinglinrain:
I’d also like to clarify the use of Claude in this PR.

This PR was implemented and submitted with Claude as a coding collaborator, while I reviewed the changes and verified the implementation and tests myself.

In the initial commit, I included Claude as a `Co-authored-by` contributor. When I amended the commit and force-pushed the updated history, I removed the `Co-authored-by` line.

Would the project be okay with Claude being listed as a co-author for this PR?

ramnes:
I prefer to avoid Claude as co-author to clarify that I hold contributors (and myself) responsible for the code that is added to this repository, not Claude. Claude is just a tool. But it also doesn't matter much and I wouldn't block a PR because Claude is listed as a co-author.

ramnes (approved):

ramnes:
Merged, thanks!

PR #396, opened 2026-09-10
Recognize `invalid_beta` API error code

tsinglinrain:
## Description

The Notion API introduced an opt-in `Notion-Beta` request header and a matching `invalid_beta` error code (HTTP 400), returned when the header is malformed, unknown, duplicated, or references a stale revision.

Before this change, a response body with `"code": "invalid_beta"` was not a known `APIErrorCode`, so `build_request_error` fell back to `UnknownHTTPResponseError` and the `code`, `additional_data`, and `request_id` fields from the body were dropped. After this change the response is parsed as an `APIResponseError` with `error.code == APIErrorCode.InvalidBeta`, and `is_http_response_error` / `APIResponseError.is_api_response_error` recognize it.

Aligns with makenotion/notion-sdk-js#754.

### DIFF
--- a/notion_client/errors.py
+++ b/notion_client/errors.py
@@ -44,6 +44,10 @@ class APIErrorCode(str, Enum):
     InvalidRequest = "invalid_request"
     """This request is not supported."""
 
+    InvalidBeta = "invalid_beta"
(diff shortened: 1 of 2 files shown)


ramnes (approved):

ramnes:
Thanks again :)

PR #398, opened 2026-09-12
Add `ray_id` and edge-proxy diagnostics to HTTP errors

tsinglinrain:
## Description

> When a request is rejected at the network edge, the response may be HTML rather than a Notion error body. The SDK already falls back to `UnknownHTTPResponseError`, but the generic message and raw HTML do not make the failure easy to diagnose.

## Deviation from the JS port

Unlike the JS port, the failure log reads both identifiers through one generic helper. The private BaseClient._extract_request_id(obj) is renamed and generalized to BaseClient._extract_str_field(obj, name); the body is unchanged apart from taking the field name as an argument. This keeps both identifiers on one code path instead of adding an `is_http_response_error` guard in a log formatter purely to read an attribute, and it lets the two logging helpers share one extractor. 

The method is private and had no callers outside `client.py`, so the change is not a public API change.

Aligns with makenotion/notion-sdk-js#751.

### DIFF
--- a/notion_client/client.py
+++ b/notion_client/client.py
@@ -217,17 +217,17 @@ class BaseClient:
 
         return response.json()
 
-    def _extract_request_id(self, obj: Any) -> Optional[str]:
-        """Extracts request_id from an object if present."""
+    def _extract_str_field(self, obj: Any, name: str) -> Optional[str]:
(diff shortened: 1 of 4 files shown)


ramnes (approved):

ramnes:
Thanks you!

PR #399, opened 2026-09-12
Add `create` method to `blocks.meeting_notes`

tsinglinrain:
## Description

`BlocksMeetingNotesEndpoint` only exposed `query()`, so creating a meeting note required calling `client.request()` directly. This adds `client.blocks.meeting_notes.create()`, which sends a `POST` to `blocks/meeting_notes` with the `title`, `language`, `options`, `source` and `parent` body parameters.

`source` is either a file upload (`{"type": "file_upload", "file_upload_id": ...}`, which requires `parent`) or an existing audio, video or file block (`{"type": "block", "block_id": ...}`, which takes no `parent`). As everywhere else in this SDK, the parameters are passed through as-is and validated by the API rather than by the client.

Aligns with makenotion/notion-sdk-js#753.

### DIFF
--- a/notion_client/api_endpoints.py
+++ b/notion_client/api_endpoints.py
@@ -41,6 +41,18 @@ class BlocksChildrenEndpoint(Endpoint):
 
 
 class BlocksMeetingNotesEndpoint(Endpoint):
+    def create(self, **kwargs: Any) -> SyncAsync[Any]:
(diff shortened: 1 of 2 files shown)


ramnes (approved):

ramnes:
Merged, thanks!

